PR #43, opened 2025-11-21
Improve `/v1/models` endpoint

sr-tream:
- fix the `type` field - it's must be a model type, like 'chat' or 'embeddings'
- add field `context_length` - its used by most clients, what supports reading context size from `/v1/models`

### DIFF
--- a/src/routes/chat-completions/handler.ts
+++ b/src/routes/chat-completions/handler.ts
@@ -44,7 +44,7 @@ export async function handleCompletion(c: Context) {
   if (isNullish(payload.max_tokens)) {
     payload = {
       ...payload,
-      max_tokens: selectedModel?.capabilities.limits.max_output_tokens,
+      max_tokens: selectedModel?.capabilities.limits?.max_output_tokens,
     }
     logger.debug("Set max_tokens to:", JSON.stringify(payload.max_tokens))
   }
--- a/src/routes/models/route.ts
+++ b/src/routes/models/route.ts
@@ -16,11 +16,14 @@ modelRoutes.get("/", async (c) => {
     const models = state.models?.data.map((model) => ({
       id: model.id,
       object: "model",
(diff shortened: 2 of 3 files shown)


caozhiyuan:
@sr-tream What is the usage scenario for context_length, and it seems that /v1/models does not have a standard definition of which fields should be included?

sr-tream:
> it seems that /v1/models does not have a standard definition of which fields should be included

It's true, but 3rd-party providers with OpenAI Compatible endpoints used field `context_length` to share context size, and some clients can parse it.

Providers what used this field (not all): SubModel, DeepInfra (but DI pass it under `metadata` object), Fireworks AI.

Part of fireworks output for example:
```json
{
"id": "accounts/fireworks/models/qwen3-vl-235b-a22b-instruct",
"object": "model",
"owned_by": "fireworks",
"created": 1758745689,
"kind": "HF_BASE_MODEL",
"supports_chat": true,
"supports_image_input": true,
"supports_tools": true,
"context_length": 262144
},
{
"id": "accounts/fireworks/models/glm-4p5",
"object": "model",
"owned_by": "fireworks",
"created": 1753809636,
"kind": "HF_BASE_MODEL",
"supports_chat": true,
"supports_image_input": false,
"supports_tools": true,
"context_length": 131072
},
{
"id": "accounts/revolut/models/cross-sell-product-selection-qwen3-coder-30b-v3",
"object": "model",
"owned_by": "revolut",
"created": 1758757208,
"kind": "HF_PEFT_ADDON",
"supports_chat": true,
"supports_image_input": false,
"supports_tools": false
}
```

caozhiyuan:
@sr-tream Besides context_length, the maximum output token parameter is also important. If it doesn't affect usage, we will temporarily leave it unchanged. It is recommended to directly submit this PR to the main repository.

PR #55, opened 2025-12-10
fix: improve thinking block handling for signature-only responses

dudufcb1:
## Summary
- Refactor `handleAssistantMessage` to conditionally include reasoning fields only when thinking content exists
- Add model-based conditional logic for thinking blocks:
  - **Claude models**: Skip empty thinking blocks (signature-only) since Claude handles signature internally
  - **Non-Claude models** (Gemini/GPT): Keep signature-only thinking blocks for conversation continuity

## Test plan
- [ ] Test with Claude Code using extended thinking on Claude models
- [ ] Verify thinking blocks are properly translated when only signature is present on non-Claude models

### DIFF
--- a/src/routes/messages/non-stream-translation.ts
+++ b/src/routes/messages/non-stream-translation.ts
@@ -143,40 +143,44 @@ function handleAssistantMessage(
     (block): block is AnthropicThinkingBlock => block.type === "thinking",
   )
 
-  const allThinkingContent = thinkingBlocks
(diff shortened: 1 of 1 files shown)


renderpxxx:
goat

caozhiyuan:
@dudufcb1 please retry in all brach , now claude model also support thinking , you should toggle think on in claude code.

PR #64, opened 2026-01-01
fix: handle signature incompatibility when resuming cross-API sessions

### LINKED ISSUE #63: Invalid signature in thinking block
When I tried to resume a session that was run with claude API (not copilot api) with copilot-api, I met
```
ERROR  HTTP error: { error:                                                                                                                                        11:49:13 AM
   { message: 'messages.1.content.0: Invalid signature in thinking block',
     code: 'invalid_request_body' } }
```

Note that it didn't happen from master branch.

## Steps to reproduce
Initiate new session with claude code using claude API (or claude subscription), then resume that session with copilot-api.

HyunggyuJang:
## Problem

When resuming a session created with Claude API using Copilot API (or vice versa), thinking block signatures are incompatible and cause errors:
```
ERROR  HTTP error: { error:
   { message: 'messages.1.content.0: Invalid signature in thinking block',
     code: 'invalid_request_body' } }
```

## Root Cause

Thinking block signatures are model-specific and tied to the exact model name:
- **Claude API (subscription)**: Uses `claude-sonnet-4-5-20250929`
- **GitHub Copilot API**: Uses `claude-sonnet-4.5`

When resuming a session created with one API using the other API, the signature was generated with a different model name, making it invalid for the target API.

## Solution

Implemented automatic retry with signature stripping:
1. Try request normally
2. If signature error detected, strip all thinking block signatures
3. Retry request with signatures removed
4. Preserve thinking text content (only remove incompatible cryptographic signatures)

## Changes

- ✅ Add automatic retry on /v1/messages endpoint for signature errors
- ✅ Gracefully strip incompatible signatures while preserving thinking text
- ✅ Suppress expected error logging for signature incompatibility
- ✅ Extract retry logic into reusable functions (DRY)
- ✅ Move error logging from service layer to handler layer for consistency

## Testing

Tested with real session resumption:
- ✅ Session created with Claude API → resumed with Copilot API works
- ✅ Normal sessions (same API) continue working without overhead
- ✅ Both streaming and non-streaming requests work
- ✅ Both Chat Completions and Responses API paths work

Fixes #63

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
         run: bun install
 
       - name: Run linter
-        run: bun run lint
+        run: bun run lint:all
 
       - name: Run type check
         run: bun run typecheck
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -1,0 +1,91 @@
+name: Docker Build and Push
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+on:
+  push:
+    # branches: [ "main" ]
+    # Publish semver tags as releases.
+    tags: [ 'v*.*.*' ]
+    paths-ignore:
+      - 'docs/**'
+
+env:
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: ghcr.io
+  # github.repository as <account>/<repo>
+  #IMAGE_NAME: ${{ github.repository }}
+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      # This is used to complete the identity challenge
+      # with sigstore/fulcio when running outside of PRs.
+      id-token: write
+
+    steps:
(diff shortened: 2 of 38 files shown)


HyunggyuJang:
@caozhiyuan feel free to switch the base branch whichever you think appropriate.

HyunggyuJang:
@caozhiyuan Turned out after changing the base to feature/chat-completions-reasoning and refactoring, the behavior has been changed, I'll work on tonight (on my timezone) since it gets messy.

HyunggyuJang:
After thorough investigation, I've discovered that this PR appeared to work, but for the wrong reasons.

### What This PR Actually Did

The retry logic accidentally **stripped all thinking blocks** instead of fixing signature incompatibility. This "worked" by removing the problematic signatures entirely, but also lost all thinking context from the conversation.

### Why Retry-Based Approaches Don't Work

Any retry that strips `reasoning_opaque` will:
1. Strip ALL thinking block signatures (not just incompatible ones)
2. Lose thinking block verification/caching benefits
3. Still fail if the thinking block structure itself is different

Closing as this approach doesn't actually solve the problem, it only masks it by removing functionality.

Related to #63

PR #65, opened 2026-01-01
fix: use default API URL when account type is individual

HyunggyuJang:
## Problem

After recent refactoring, business users who don't explicitly specify their account type can no longer connect, breaking previous behavior.

## Solution

When account type is not specified or set to 'individual', use the default `api.githubcopilot.com` URL instead of constructing `api.individual.githubcopilot.com`.

The default URL works for both individual and business accounts, restoring the previous behavior where business users could work without the `-a business` flag.

## Changes

- Special case for `individual` account type to use default URL
- Business/enterprise account types still use their specific subdomains when explicitly set
- Maintains backward compatibility

## Testing

- ✅ Individual accounts work with default URL
- ✅ Business accounts work without specifying account type (previous behavior restored)
- ✅ Business/enterprise accounts continue to work when explicitly specified

### DIFF
--- a/src/lib/api-config.ts
+++ b/src/lib/api-config.ts
@@ -14,8 +14,9 @@ const USER_AGENT = `GitHubCopilotChat/${COPILOT_VERSION}`
 const API_VERSION = "2025-10-01"
 
 export const copilotBaseUrl = (state: State) =>
-  `https://api.${state.accountType}.githubcopilot.com`
-
+  state.accountType === "individual" ?
+    "https://api.githubcopilot.com"
+  : `https://api.${state.accountType}.githubcopilot.com`
 export const copilotHeaders = (state: State, vision: boolean = false) => {
   const headers: Record<string, string> = {
     Authorization: `Bearer ${state.copilotToken}`,


caozhiyuan:
@HyunggyuJang  https://api.individual.githubcopilot.com worked for me , Can you open https://api.individual.githubcopilot.com in your browser?

caozhiyuan:
In the vscode github copilot extension, it also calls https://api.individual.githubcopilot.com, so I simplified the domain name @HyunggyuJang

caozhiyuan:
@dudufcb1 https://api.individual.githubcopilot.com/ worked for you?  It seems that in some places, this domain cannot be accessed?

caozhiyuan:
@HyunggyuJang you are not individual plan?

dudufcb1:
> https://api.individual.githubcopilot.com/

is for me, but is not acceded from the browser as its post endpoint. Im working with the default version.

HyunggyuJang:
> @HyunggyuJang https://api.individual.githubcopilot.com worked for me , Can you open https://api.individual.githubcopilot.com in your browser?

@caozhiyuan 
It didn’t work for me, and might not work for others too. If the default version works for everyone, why not going with it?

caozhiyuan:
> > @HyunggyuJang https://api.individual.githubcopilot.com worked for me , Can you open https://api.individual.githubcopilot.com in your browser?
> 
> @caozhiyuan 
> It didn’t work for me, and might not work for others too. If the default version works for everyone, why not going with it?  

@HyunggyuJang maybe you are using business account type? can pr for chat completions reasoning branch? and pr commit message should correct , for individual plan , it worked ok.

HyunggyuJang:
@caozhiyuan Sorry to this noise! Turned out I mangled the config so that my individual account wasn’t set up for the current machine (I use both individual, business plans). I’ll try with correct account type and get back to you. Once I confirmed that current account type handling works fine, I’ll close this PR. Thanks for your clear guidance and sorry for this disturbance.

caozhiyuan:
> @caozhiyuan Sorry to this noise! Turned out I mangled the config so that my individual account wasn’t set up for the current machine (I use both individual, business plans). I’ll try with correct account type and get back to you. Once I confirmed that current account type handling works fine, I’ll close this PR. Thanks for your clear guidance and sorry for this disturbance.

@HyunggyuJang welcome to pr for chat completions reasoning branch, seems all account type can use api.githubcopilot.com . and should commit correct message, to prevent others from being unaware of the reason for the changes later.

HyunggyuJang:
> @HyunggyuJang welcome to pr for chat completions reasoning branch, seems all account type can use api.githubcopilot.com . and should commit correct message, to prevent others from being unaware of the reason for the changes later.

@caozhiyuan Sorry, I cannot understand the commit message part. Did I something wrong? It seems okay to me (although it did assume the problem wrongly).

caozhiyuan:
@HyunggyuJang Do you currently need to specify the account type? You can check the commit history. Previously, I simplified the baseurl logic, but it seems to have affected business account users, who previously could work without specifically setting the account type. You can change the commit message to: by default, use api.githubcopilot.com for all account types

HyunggyuJang:
> @HyunggyuJang Do you currently need to specify the account type? You can check the commit history. Previously, I simplified the baseurl logic, but it seems to have affected business account users, who previously could work without specifically setting the account type. You can change the commit message to: by default, use api.githubcopilot.com for all account types

I got it, I thought I will drop the commit since my assumption was wrong, but you are suggesting to change the direction to use default API server URL instead of handling account types at all?

HyunggyuJang:
@caozhiyuan If you want to explore that direction, I will change the base branch and refactor not to handle account types at all and use default API URL, which will need more modifications.

In either way, I think would be better close this PR as doesn't fit with the original purpose.

caozhiyuan:
@HyunggyuJang The previously submitted code is fine, only the commit message is wrong, just need to correct the message.

HyunggyuJang:
@caozhiyuan Got it, then please clarify the purpose. Do you want to preserve previous behavior (business user can use it without account type option, which has been changed due to your refactoring) or want to discard account type management at all?

caozhiyuan:
@HyunggyuJang preserve previous behavior (business user can use it without account type option, which has been changed due to your refactoring)

HyunggyuJang:
@caozhiyuan Got it, will change commit message and PR title & description.

HyunggyuJang:
@caozhiyuan Fixed, PTAL

PR #97, opened 2026-02-26
fix: correct dev and start script commands in package.json

sup3rgiu:
- Running `bun run start` or `bun run dev` fails with "No command specified"
- Add `start` subcommand to both scripts so they work again


caozhiyuan:
@sup3rgiu can  `bun run start start -v`  or `bun run start auth`

PR #101, opened 2026-03-02
fix: prevent premium request overcounting and model name mismatch for Claude Agent Teams feature

### LINKED ISSUE #94: Failed when using claude code agent team mode
When runnign with `claude --teammate-mode tmux` and agent team started. Error will be shown as follow:

Error from Claude Code
```
API Error: 400 {"error":{"message":"{\"error\":{\"message\":\"The requested model is not
     supported.\",\"code\":\"model_not_supported\",\"param\":\"model\",\"type\":\"invalid_request_error\"}}\n","type":"error"}}
```

Error from copilot-api
```
Error occurred: Failed to create chat completions 
    at createChatCompletions (/Users/username/.npm/_npx/f938ad9b3827e216/node_modules/@jeffreycao/copilot-api/dist/server-DbxsVRo3.js:439:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async handleWithChatCompletions (/Users/username/.npm/_npx/f938ad9b3827e216/node_modules/@jeffreycao/copilot-api/dist/server-DbxsVRo3.js:1947:19)
    at async handleCompletion (/Users/username/.npm/_npx/f938ad9b3827e216/node_modules/@jeffreycao/copilot-api/dist/server-DbxsVRo3.js:1940:9)
    at async /Users/username/.npm/_npx/f938ad9b3827e216/node_modules/@jeffreycao/copilot-api/dist/server-DbxsVRo3.js:2130:10
    at async dispatch (/Users/username/.npm/_npx/f938ad9b3827e216/node_modules/hono/dist/compose.js:22:17)
    at async dispatch (/Users/username/.npm/_npx/f938ad9b3827e216/node_modules/hono/dist/compose.js:22:17)
    at async cors2 (/Users/username/.npm/_npx/f938ad9b3827e216/node_modules/hono/dist/middleware/cors/index.js:79:5)
    at async dispatch (/Users/username/.npm/_npx/f938ad9b3827e216/node_modules/hono/dist/compose.js:22:17)
    at async logger2 (/Users/username/.npm/_npx/f938ad9b3827e216/node_modules/hono/dist/middleware/logger/index.js:38:5)
```

sandriaas:
## Problem

  Two issues when Agent Teams are active:

  **1. Excessive premium request consumption**

  The team_lead accumulates premium requests far beyond what the user actually
  initiated. Every teammate notification, idle poll, and system reminder arrives
  as a plain text user message — and the proxy's `isInitiateRequest` heuristic
  in `create-messages.ts` treats any non-tool_result user message as
  user-initiated (`X-Initiator: user`).

  Teammates are fine because they carry the subagent marker → always
  `X-Initiator: agent`. The team_lead has no marker, so it falls back to the
  heuristic and every injected system message burns a premium request.

  **2. Wrong model name sent under teammates**

  When `ANTHROPIC_DEFAULT_OPUS_MODEL` is configured (e.g. `claude-sonnet-4.6`),
  requests under teammates still go out as `claude-opus-4-6` (dot-separated
  version suffix). Claude Code sends the full versioned model ID instead of the
  base model name in teammate context, causing the proxy to not recognize it.

  Fixes #94

  ## Fix

  **Premium request fix:** After `mergeToolResultForClaude`, detect
  system-generated text-only user messages by pattern-matching known tags:

  - `<system-reminder>`
  - `<teammate-message>`
  - `<local-command-caveat>`
  - JSON protocol messages (`idle_notification`, `shutdown*`, `teammate_terminated`)

  If every text block in the last user message matches one of these patterns,
  override the initiator to `"agent"` before the request goes out. Only triggers
  when there's no subagent marker (team_lead only). Genuine user messages won't
  match any pattern and stay `X-Initiator: user`.

  **Model name fix:** `normalizeModelName` strips dot-separated version suffixes
  so `claude-opus-4-6` and `claude-opus-4.6` both resolve to `claude-opus-4`,
  allowing the proxy to correctly identify the model regardless of how Claude
  Code formats the name.

  Also fixes pre-existing `no-useless-escape` lint warnings in `normalizeModelName`.

  ## Testing

  Ran a 5+ agent swarm — premium request count only increased by ~3
  (the actual user-initiated turns) instead of spiking on every teammate
  notification.

### DIFF
--- a/src/routes/messages/handler.ts
+++ b/src/routes/messages/handler.ts
@@ -50,6 +50,39 @@ import { parseSubagentMarkerFromFirstUser } from "./subagent-marker"
 
 const logger = createHandlerLogger("messages-handler")
 
+const SYSTEM_GENERATED_PATTERNS = [
+  /^<system-reminder>/,
+  /^<teammate-message\s/,
+  /^<local-command-caveat>/,
+  /^\{"type":"idle_notification"/,
+  /^\{"type":"shutdown/,
+  /^\{"type":"teammate_terminated"/,
+]
+
+const isSystemGeneratedText = (text: string): boolean => {
+  const trimmed = text.trim()
+  return SYSTEM_GENERATED_PATTERNS.some((pattern) => pattern.test(trimmed))
+}
+
+const detectSystemGeneratedInitiator = (
+  payload: AnthropicMessagesPayload,
+): "agent" | undefined => {
+  const lastUserMsg = payload.messages.at(-1)
+  if (
+    lastUserMsg?.role === "user"
+    && Array.isArray(lastUserMsg.content)
+    && lastUserMsg.content.length > 0
+    && lastUserMsg.content.every(
+      (block) =>
+        block.type === "tool_result"
+        || (block.type === "text" && isSystemGeneratedText(block.text)),
+    )
+  ) {
+    return "agent"
+  }
+  return undefined
+}
+
 const compactSystemPromptStart =
   "You are a helpful AI assistant tasked with summarizing conversations"
 
@@ -60,7 +93,8 @@ export async function handleCompletion(c: Context) {
   logger.debug("Anthropic request payload:", JSON.stringify(anthropicPayload))
 
   const subagentMarker = parseSubagentMarkerFromFirstUser(anthropicPayload)
-  const initiatorOverride = subagentMarker ? "agent" : undefined
+  let initiatorOverride: "agent" | "user" | undefined =
+    subagentMarker ? "agent" : undefined
   if (subagentMarker) {
     logger.debug("Detected Subagent marker:", JSON.stringify(subagentMarker))
   }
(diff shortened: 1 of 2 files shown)


sandriaas:
## Update: Added model name normalization fix

The initial commit added `_proxy_bypass: "subagent"` to the marker for robust detection of `team_lead`/`teammate` agent types. After further investigation, I identified the **root cause** of the `400 model_not_supported` error:

### Root Cause
Copilot's model proxy uses **dot-separated** version IDs (e.g. `claude-opus-4.6`), but `translateModelName` in `non-stream-translation.ts` only matched **dash-separated** versions (`claude-opus-4-*`). This means:

1. `claude-opus-4.6` fails the `startsWith("claude-opus-4-")` check
2. The model name is sent as-is to the upstream API
3. Upstream rejects it with `400 model_not_supported`

Additionally, the model lookup in `handler.ts` uses **exact match** against `state.models.data`. When the versioned model (e.g. `claude-opus-4.6`) doesn't match the base model entry (`claude-opus-4`), `selectedModel` is `undefined`, causing the request to fall through to chat completions instead of the correct Messages/Responses API path.

### Fix (second commit)
- Updated `translateModelName` regex to handle both `[.-]` separators
- Added a **normalized fallback** to the model lookup in `handler.ts` so versioned models route to the correct API path

Both commits together address:
1. **Marker detection** — `_proxy_bypass` ensures new agent types are recognized
2. **Model routing** — dot-separated versions are properly normalized

sandriaas:
Pushed a new commit that fixes the premium request overcounting for Agent Teams team_lead.

**Root cause:** When Agent Teams are active, the team_lead receives teammate notifications, idle notifications, and system reminders as text-only user messages. The existing `isInitiateRequest` heuristic in `create-messages.ts` treats these as user-initiated (`X-Initiator: user`), incorrectly consuming premium requests. Teammates are fine (they have the subagent marker → always `X-Initiator: agent`), but the team_lead has no marker.

**Fix:** After `mergeToolResultForClaude`, detect system-generated text-only user messages by pattern-matching known tags (`<system-reminder>`, `<teammate-message>`, `<local-command-caveat>`, and JSON protocol messages like `idle_notification`, `shutdown`, `teammate_terminated`). When all text blocks in the last user message are system-generated, override the initiator to `"agent"`.

**Safety:**
- Only triggers when there's no subagent marker (team_lead only)
- Only triggers when ALL text blocks are system-generated
- Genuine user-initiated messages won't match → stays `X-Initiator: user`

**Tested:** 5+ agent team swarm — premium requests only increased by ~3 (initial user messages) instead of on every teammate notification.

Also fixed pre-existing `no-useless-escape` lint warnings in `normalizeModelName`.

caozhiyuan:
For now, let's not merge it. The current modification carries a risk of being banned, and we need to wait until the teammate feature stabilizes before deciding how to proceed.

PR #103, opened 2026-03-04
fix: filter unsupported context-1m beta header

RoyShen12:
## Summary

- Filter out `context-1m-*` prefixed beta headers when forwarding requests to the upstream Copilot API
- Claude Code sends `context-1m-2025-08-07` in the `anthropic-beta` header when using 1M context models (e.g., `claude-opus-4.6-1m`), but the upstream API does not support this, causing **400 errors** with `unsupported beta header(s): context-1m-2025-08-07`
- Uses `startsWith("context-1m-")` prefix matching to future-proof against date suffix changes

## Error before fix

```
ERROR  HTTP error: { error:
   { message: 'unsupported beta header(s): context-1m-2025-08-07',
     code: 'invalid_request_body' } }
```

## Test plan

- [ ] Verify requests with `context-1m-2025-08-07` beta header no longer return 400
- [ ] Verify other valid beta headers are still forwarded correctly
- [ ] Verify thinking mode fallback (`interleaved-thinking-2025-05-14`) still works when no beta header is provided

### DIFF
--- a/src/services/copilot/create-messages.ts
+++ b/src/services/copilot/create-messages.ts
@@ -48,7 +48,10 @@ export const createMessages = async (
     const filteredBeta = anthropicBetaHeader
       .split(",")
       .map((item) => item.trim())
-      .filter((item) => item !== "claude-code-20250219")
+      .filter(
+        (item) =>
+          item !== "claude-code-20250219" && !item.startsWith("context-1m-"),
+      )
       .join(",")
     if (filteredBeta) {
       headers["anthropic-beta"] = filteredBeta


caozhiyuan:
@RoyShen12 fixed , align with vscode copilot extension anthropic-beta

const INTERLEAVED_THINKING_BETA = "interleaved-thinking-2025-05-14"
const allowedAnthropicBetas = new Set([
  INTERLEAVED_THINKING_BETA,
  "context-management-2025-06-27",
  "advanced-tool-use-2025-11-20",
])

dylbarne:
if potentially related, found that am running into error:
```
 ERROR  HTTP error: { type: 'error', 
  error:
   { type: 'invalid_request_error',
     message: 'prompt is too long: 4194174 tokens > 1000000 maximum' }
```
     
when setting `/model` to opus, it set as `"model": "opus[1m]"` in settings.json
     
I thought maybe setting env var `CLAUDE_CODE_DISABLE_1M_CONTEXT = 1` , from [docs](https://code.claude.com/docs/en/model-config#extended-context) may fix this, but it did not
(seems to also be the case for Sonnet as well)

@caozhiyuan any suggestions?   (I can also open up a separate new issue if preferred)

caozhiyuan:
Github Copilot does not explicitly support 1M context, although you can use it illegally, but be careful to be banned. Just select 200K in length in the Claude Code. In the long context, the AI will be unstable. @dylbarne

dylbarne:
> Github Copilot does not explicitly support 1M context, although you can use it illegally, but be careful to be banned. Just select 200K in length in the Claude Code. In the long context, the AI will be unstable. @dylbarne

thank you for your reply

I am using only ClaudeCode settings.json (not opencode), is there a way to set these?

I only really saw related was Max Output Tokens `CLAUDE_CODE_MAX_OUTPUT_TOKENS` in [docs](https://code.claude.com/docs/en/env-vars#environment-variables), what are you recommending to set exactly to fix this?

(note: I also feel like I did not have this issue at all ~yesterday or ealier, maybe also has to do with something GitHub Copilot changed on their endpoints backend)

caozhiyuan:
/model selects a model without 1M ending @dylbarne

dylbarne:
> /model selects a model without 1M ending @dylbarne

yes I tried that, did not seem to work

on second thought, I feel like maybe is due to some change in Claude Code itself (around 1M updates, potentially starting [here](https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md#2175))

I will try to downgrade to lower version of Claude Code to verify

edit: nevermind , tried `2.1.47` and still hitting same issue

caozhiyuan:
@dylbarne 2.1.79 config like this 
    "ANTHROPIC_MODEL": "claude-sonnet-4-6",
    "ANTHROPIC_DEFAULT_SONNET_MODEL": "claude-sonnet-4-6",
 is ok . 
2.1.77 can select a model without 1M ending . 
you can run /context check it

caozhiyuan:
@dylbarne In fact, GitHub Copilot has supported a 1M context since Sonnet 4.5, and it won't report an error even if you exceed 200K. However, exceeding 200K would definitely violate the GitHub Copilot agreement.

dylbarne:
> @dylbarne 2.1.79 config like this "ANTHROPIC_MODEL": "claude-sonnet-4-6", "ANTHROPIC_DEFAULT_SONNET_MODEL": "claude-sonnet-4-6", is ok . 2.1.77 can select a model without 1M ending . you can run /context check it

ah ok I think I see the issue, looking at `/context`

for a particular repo that I was hitting this on, is automatically, Context Usage:
```
claude-opus-4.6 · 2885k/200k tokens (1443%)
Estimated usage by category
System prompt: 3.4k tokens (1.7%)
System tools: 2880.8k tokens (1440.4%)
Memory files: 512 tokens (0.3%)
Skills: 417 tokens (0.2%)
Autocompact buffer: 33k tokens (16.5%)
```

I have no idea why System tools is so crazy for this repo/directory in particular, ~~but I will investigate~~ 
_found the cause: seems like is from their `sandbox` setting I was messing around with in particular repo... gets brought into System Tools for whatever reason... disabled Sandbox and context usage returned to normal_

caozhiyuan:
@dylbarne a lot of skills and mcp?

dylbarne:
> @dylbarne a lot of skills and mcp?

was something related to their Sandbox settings (updated my previous response)

only CC issue I could find that seemed related was [this](https://github.com/anthropics/claude-code/issues/27757) really

I just disabled sandbox related functionality and returned back to normal though


thank you for your help and continued support of this project as always!

PR #111, opened 2026-03-07
fix: preserve Claude thinking blocks for Claude Code

mouadlotfi:
## Summary
- fix Claude-model thinking translation so Claude Code no longer receives malformed placeholder or signature-only thinking blocks from Copilot reasoning streams
- buffer Claude streaming output until the reasoning signature arrives, and align the non-streaming path with the same behavior
- add safe thinking telemetry plus regression coverage for Claude reasoning compatibility

## Verification
- bun run lint
- bun test

### DIFF
--- a/src/routes/messages/anthropic-types.ts
+++ b/src/routes/messages/anthropic-types.ts
@@ -1,3 +1,5 @@
+import type { ChatCompletionChunk } from "~/services/copilot/create-chat-completions"
+
 // Anthropic API Types
 
 export interface AnthropicMessagesPayload {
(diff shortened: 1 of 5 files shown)


caozhiyuan:
@mouadlotfi Claude model no longer uses the chat-completions interface, is yours an enterprise version?

claude model use message api `https://github.com/caozhiyuan/copilot-api/blob/all/src/routes/messages/handler.ts#L103`

caozhiyuan:
You can run bun run start start -v, logs are located at ~/.local/share/copilot-api/logs, no need to print logs repeatedly. If you want to submit a PR, please provide the corresponding log files for the issue background; currently I don't know what the problem is.

mouadlotfi:
@caozhiyuan Thanks for explaining. You're right, I didn't realize Claude was going through the Messages API path where thinking already works fine. I was confused and thought it was broken on my end.

PR #120, opened 2026-03-13
Better support for Copilot model names

gingi:


caozhiyuan:
@gingi supported. Currently, the Claude model is using the message API.

PR #137, opened 2026-03-24
fix: add fetch retry with connection pool reset on network errors

lovinrain:
## Summary

Fixes recurring `ENETUNREACH` / `ECONNREFUSED` errors when connecting to GitHub's Copilot API that previously required a full server restart to resolve.

## Root Cause

Node.js's `fetch()` is backed by **undici**, which maintains a persistent HTTP connection pool via a global `Agent`. The default Agent has **no keep-alive timeout**, so once a TCP connection to a GitHub IP (e.g. `140.82.113.22`) is established, it stays in the pool indefinitely.

When GitHub rotates IPs or a network route changes, the cached connection becomes stale but undici keeps trying to reuse it — resulting in `ENETUNREACH`. Every subsequent request fails because they all go through the same stale pool. Restarting the server works because it destroys the process (and the pool), forcing fresh DNS resolution and new connections.

This was especially problematic without the `--proxy-env` flag, because in that case **no custom Agent was configured at all** — the server ran on undici's bare defaults.

## Fix

Two-pronged approach:

### 1. Global Agent with proper timeouts (`src/lib/proxy.ts`)

- **`keepAliveTimeout: 30s`** — idle connections are automatically evicted after 30 seconds, preventing stale connections from lingering
- **`keepAliveMaxTimeout: 60s`** — hard upper bound on keep-alive
- **`connect.timeout: 15s`** — connection attempts don't hang forever
- `initDefaultAgent()` is now called **unconditionally** at startup (not gated behind `--proxy-env`), ensuring every server instance gets a properly configured Agent
- `resetAgent()` function that closes the old Agent/pool and creates a fresh one — works in both direct and proxy modes

### 2. `fetchWithRetry` wrapper (`src/lib/fetch-retry.ts`)

A drop-in replacement for `fetch()` that retries on network-level errors:

- **Detects network errors**: `ENETUNREACH`, `ECONNREFUSED`, `ECONNRESET`, `EPIPE`, `ETIMEDOUT`, `UND_ERR_SOCKET`, `UND_ERR_CONNECT_TIMEOUT`
- **Max 2 retries** (3 total attempts) with delays of 500ms and 1500ms
- **Resets the connection pool** on first retry via `resetAgent()` — this is effectively what a restart does, but without restarting
- **Only retries network errors** — HTTP errors (4xx/5xx) are never retried, they pass through as before
- **Safe for streaming**: retry only triggers when `fetch()` itself throws (before any Response body exists)

### Call sites updated (9 files)

All copilot and GitHub service fetch calls now use `fetchWithRetry`:
- `create-messages.ts`, `create-responses.ts`, `create-chat-completions.ts`, `create-embeddings.ts`, `get-models.ts`
- `get-copilot-token.ts`, `get-copilot-usage.ts`, `get-user.ts`, `get-device-code.ts`

### Excluded (already have their own error handling)
- `poll-access-token.ts` — has its own retry loop
- `anthropic-proxy.ts` — external provider, not affected by our connection pool
- `count-tokens-handler.ts` — already falls back to local estimation on failure

## Test plan

- [x] `npm run lint` passes
- [x] `npm run build` passes
- [x] Type-check passes (no new errors in modified files)
- [ ] Manual test: start server, make requests, verify normal operation
- [ ] Monitor logs for retry messages (`Network error (attempt X/3), retrying...`) during transient network issues

### DIFF
--- a/src/lib/fetch-retry.ts
+++ b/src/lib/fetch-retry.ts
@@ -1,0 +1,56 @@
+import consola from "consola"
+
+import { resetAgent } from "./proxy"
+
+const RETRYABLE_CODES = new Set([
+  "ENETUNREACH",
+  "ECONNREFUSED",
+  "ECONNRESET",
+  "EPIPE",
+  "ETIMEDOUT",
+  "UND_ERR_SOCKET",
+  "UND_ERR_CONNECT_TIMEOUT",
+])
+
+const MAX_RETRIES = 2
+const RETRY_DELAYS = [500, 1500]
+
+function isNetworkError(error: unknown): boolean {
+  if (!(error instanceof TypeError)) return false
+  const cause = (error as { cause?: { code?: string } }).cause
+  if (cause?.code && RETRYABLE_CODES.has(cause.code)) return true
+  if (error.message === "fetch failed") return true
+  return false
+}
+
+export async function fetchWithRetry(
+  input: string | URL | Request,
+  init?: RequestInit,
+): Promise<Response> {
+  let lastError: unknown
+
+  for (let attempt = 0; attempt <= MAX_RETRIES; attempt++) {
+    try {
+      return await fetch(input, init)
+    } catch (error) {
+      lastError = error
+      if (!isNetworkError(error) || attempt === MAX_RETRIES) {
+        throw error
+      }
+
+      const causeCode =
+        (error as { cause?: { code?: string } }).cause?.code ?? ""
+      consola.warn(
+        `Network error (attempt ${attempt + 1}/${MAX_RETRIES + 1}), retrying in ${RETRY_DELAYS[attempt]}ms: ${(error as Error).message} ${causeCode}`,
+      )
+
+      if (attempt === 0) {
+        resetAgent()
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, RETRY_DELAYS[attempt]))
+    }
+  }
+
+  throw lastError
+}
--- a/src/lib/proxy.ts
+++ b/src/lib/proxy.ts
@@ -2,65 +2,114 @@ import consola from "consola"
 import { getProxyForUrl } from "proxy-from-env"
 import { Agent, ProxyAgent, setGlobalDispatcher, type Dispatcher } from "undici"
 
-export function initProxyFromEnv(): void {
-  if (typeof Bun !== "undefined") return
+const AGENT_OPTIONS: Agent.Options = {
+  keepAliveTimeout: 30_000,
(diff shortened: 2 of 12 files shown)


caozhiyuan:
I dug into the corresponding undici versions because the PR description makes a very specific claim about default behavior.

On a local check against the current project/runtime (`Node v22.21.1`, bundled `undici 6.22.0`, repo-installed `undici 7.16.0`), I don't think the stated root cause is established.

Relevant source facts:
- In `nodejs/undici@v6.22.0` and `nodejs/undici@v7.16.0`, `lib/dispatcher/client.js` sets:
  - `keepAliveTimeout == null ? 4e3 : keepAliveTimeout`
  - `keepAliveMaxTimeout == null ? 600e3 : keepAliveMaxTimeout`
- In `nodejs/undici@v6.22.0` and `nodejs/undici@v7.16.0`, `lib/core/connect.js` sets:
  - `timeout = timeout == null ? 10e3 : timeout`

So the default behavior is already finite: 4s idle keep-alive and 10s connect timeout, not "no keep-alive timeout" / "connect can hang forever".

I'm also skeptical of the "same stale pool keeps getting reused until restart" part:
- in `nodejs/undici@v6.22.0`, `lib/dispatcher/pool.js` removes clients on `connectionError` and explicitly says `They will not be re-used.`
- in `nodejs/undici@v7.16.0`, pool/agent cleanup is even more aggressive

That doesn't prove the symptom can never happen; a recovery workaround may still help in some VPN / sleep-wake / proxy / IPv6 / route-change scenarios. But based on the corresponding undici sources, this currently reads more like a mitigation/workaround than a demonstrated fix for the stated root cause.

One more reason I'm hesitant about the broad server-side retry wrapper: `Claude Code` and `opencode` already do automatic retries at the client layer. That makes it harder to justify automatic retries for non-idempotent POSTs in the server unless there is strong evidence that the server process itself gets stuck in a state client retries cannot recover from.

I think the PR description should either:
1. tone down the claim and describe this as a recovery strategy for some environments, or
2. include logs / a reproducible case showing why the existing undici eviction/reconnect behavior was insufficient and why a full process restart was required

Separate from the description, I still think automatic retries on non-idempotent POSTs are risky, because they can duplicate upstream work if the request was accepted but the response path failed.

— Drafted with GPT-5.4 xhigh.

caozhiyuan:
@lovinrain cc

caozhiyuan:
if you use vpn , try this tun mode. https://github.com/clash-verge-rev/clash-verge-rev @lovinrain

PR #162, opened 2026-04-05
feat: convert tool_reference blocks to text for proxy compatibility

gzenz:
## Summary

- Adds support for `tool_reference` content blocks inside `tool_result` messages by converting them to `text` blocks early in the request pipeline
- Claude Code's `ENABLE_TOOL_SEARCH` (MCP tool lazy-loading) returns results as `tool_reference` content blocks, which the Copilot Messages API rejects as unknown types
- Conversion runs before all other preprocessing to avoid breaking `mergeToolResultForClaude`

## Context

When `ENABLE_TOOL_SEARCH=true`, Claude Code defers MCP tool definitions and loads them on demand via a `ToolSearch` meta-tool. The results are returned as `{type: "tool_reference", tool_name: "..."}` content blocks inside `tool_result` messages. This is an Anthropic API content block type that the Copilot Messages API does not support, causing 503 errors.

The fix converts these blocks to `{type: "text", text: "[tool_reference: tool_name]"}` before the payload reaches the API. This is lossless -- Claude Code tracks discovered tools client-side and manages the `tools` array independently.

## Test plan

- [x] Verified `ENABLE_TOOL_SEARCH=true` with MCP tools (Context7, alphaxiv) through copilot-api -- no more 503 errors
- [x] Verified ToolSearch deferred tool loading works end-to-end
- [x] Build passes (`bun run build`)
- [x] Lint passes (lint-staged ran on commit)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/package.json",
   "name": "@jeffreycao/copilot-api",
-  "version": "1.5.3-beta.2",
+  "version": "1.5.3-beta.3",
   "description": "Turn GitHub Copilot into OpenAI/Anthropic API compatible server. Usable with Claude Code Or Codex Or Opencode!",
   "keywords": [
     "proxy",
--- a/src/routes/messages/anthropic-types.ts
+++ b/src/routes/messages/anthropic-types.ts
@@ -42,6 +42,11 @@ export interface AnthropicImageBlock {
   }
 }
 
+export interface AnthropicToolReferenceBlock {
+  type: "tool_reference"
+  tool_name: string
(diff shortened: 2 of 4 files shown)


caozhiyuan:
@gzenz Previously, there were no plans to support Claude code client tool search. You can see the reason in the readme settings.json. Removing the "Tool loaded..." text will solve the problem. Also, why are there so many "any" in your code!

caozhiyuan:
I'll submit the code later to make it compatible

caozhiyuan:
@gzenz supported

gzenz:
i overlooked the readme tbh. your solution also looks much better :)

PR #166, opened 2026-04-08
feat: auto-refresh Copilot token on 401 and retry request

novomanish:
## Summary

- When any of the three Copilot upstream callers (`createChatCompletions`, `createMessages`, `createResponses`) receives a **HTTP 401** response, automatically refresh the token and retry the request once before surfacing an error
- Applies uniformly to all three API paths: `/chat/completions`, `/v1/messages`, and `/responses`
- On retry, the refreshed token from `state.copilotToken` is injected into the `Authorization` header before re-sending the identical payload

## Changes

**`src/lib/with-token-refresh.ts`** (new) — shared `fetchWithCopilotTokenRefresh` helper that owns the entire refresh-and-retry flow:
- On 401: calls `setupCopilotToken()` inside a `try/catch`
  - If refresh fails → throws `HTTPError` wrapping the **original 401 response**, so callers see a controlled upstream error instead of the token-fetch failure
  - If refresh succeeds → mutates `headers["Authorization"]` in place and retries once
- All other non-ok responses → throws `HTTPError` immediately

**`createChatCompletions` / `createMessages` / `createResponses`** — each had ~20 lines of duplicated 401 handling removed; now delegate to `fetchWithCopilotTokenRefresh`.

## Motivation

Copilot tokens have a limited TTL. Long-running or idle sessions can hit 401s on otherwise valid requests. Without this change the error propagates immediately. With it, the token is refreshed transparently and the request succeeds without user intervention. The refactor also ensures consistent behavior and closes the gap where a failed refresh would short-circuit with the wrong error.

## Test plan

- [ ] Start the proxy with a valid token, wait for it to expire (or manually expire it), and verify the next request succeeds without manual intervention
- [ ] Simulate a refresh failure and confirm the caller receives an `HTTPError` reflecting the original 401, not the refresh error
- [ ] Verify streaming responses work correctly after a token refresh on all three paths (`/chat/completions`, `/v1/messages`, `/responses`)
- [ ] Confirm a genuine 401 (bad credentials) still surfaces an error after the single retry

### DIFF
--- a/src/lib/with-token-refresh.ts
+++ b/src/lib/with-token-refresh.ts
@@ -1,0 +1,53 @@
+import consola from "consola"
+
+import { HTTPError } from "~/lib/error"
+import { state } from "~/lib/state"
+import { setupCopilotToken } from "~/lib/token"
+
+/**
+ * Wraps a fetch call with automatic Copilot token refresh on 401.
+ *
+ * If the initial response is 401:
+ *   - Attempts to refresh the token via setupCopilotToken().
+ *   - If refresh fails, throws an HTTPError wrapping the original 401 response
+ *     so the caller sees a controlled upstream error, not the refresh failure.
+ *   - If refresh succeeds, mutates headers["Authorization"] and retries once.
+ *
+ * Returns a Response that is guaranteed ok, or throws HTTPError.
+ */
+export const fetchWithCopilotTokenRefresh = async (
+  doRequest: () => Promise<Response>,
+  headers: Record<string, string>,
+  errorLabel: string,
+): Promise<Response> => {
+  const response = await doRequest()
+
+  if (!response.ok) {
+    if (response.status === 401) {
+      consola.warn("Copilot token expired, refreshing and retrying...")
+      try {
+        await setupCopilotToken()
+      } catch (refreshError) {
+        consola.error(
+          `Failed to refresh Copilot token while retrying ${errorLabel}:`,
+          refreshError,
+        )
+        throw new HTTPError(`Failed to ${errorLabel}`, response)
+      }
+      headers["Authorization"] = `Bearer ${state.copilotToken}`
+      const retryResponse = await doRequest()
+      if (!retryResponse.ok) {
+        consola.error(
+          `Failed to ${errorLabel} after token refresh`,
+          retryResponse,
+        )
+        throw new HTTPError(`Failed to ${errorLabel}`, retryResponse)
+      }
+      return retryResponse
+    }
+    consola.error(`Failed to ${errorLabel}`, response)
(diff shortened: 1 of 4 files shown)


caozhiyuan:
Thanks for the improvement here. A few follow-ups before merging:

- The 401 handling is currently duplicated across `createChatCompletions`, `createMessages`, and `createResponses`. It would be cleaner and safer to extract the refresh-and-retry flow into a shared helper so the behavior stays consistent in one place.
- Please add proper error handling around `setupCopilotToken()`. If the refresh call itself fails, we currently short-circuit with the token refresh error instead of handling the original upstream 401 in a controlled way.
- Please also resolve the current lint issue/warning before merge.

caozhiyuan:
When will 401 occur? The backend has a timer to refresh the token. @novomanish

documanish:
> When will 401 occur? The backend has a timer to refresh the token. @novomanish

The use case is:
1. Claude uses proxy
2. Proxy is running in docker container
3. User put their system in standby for 24h
4. User returns and Claude stops working because API would through 401
5. User needs to restart the docker container to refresh the token

caozhiyuan:
I have two concerns with this change:

1. In `src/lib/with-token-refresh.ts`, the retry path sends a second `fetch` without draining or canceling the original 401 response body. That can leave an unread response hanging around and may tie up connections or buffers over time. If we keep this logic, we should explicitly cancel or drain the first response before retrying.

2. More broadly, I am not convinced this request-level retry is worth the added complexity right now. We already have a background token refresh loop, and when a refresh fails it retries again every 15 seconds. On top of that, concurrent 401s would each call `setupCopilotToken()` again, which increases refresh traffic and state churn. Given the existing background refresh behavior, I would lean toward keeping the request path simple and letting the refresher recover the token instead of adding another retry branch here.

caozhiyuan:
One more concern: this path is not single-flight. If several requests hit a 401 at the same time, each of them will call `setupCopilotToken()` concurrently. That means duplicate token refresh requests, extra refresh-loop churn, and more state races in a path that should ideally stay simple. Combined with the existing background refresh loop, this is another reason I would prefer not to add request-level retry logic here unless we have clear evidence that the background refresher is insufficient.

caozhiyuan:
@novomanish  cc

documanish:
makes sense.. I wonder why I had to repeatedly restart container everyday.. Let me observe the existing behavior again..

caozhiyuan:
Is your computer in sleep mode? @novomanish

documanish:
> Is your computer in sleep mode? @novomanish

yes

caozhiyuan:
After returning from sleep mode to normal mode, the first token refresh will display a network error, but the second refresh will be normal.  @documanish

PR #170, opened 2026-04-10
fix: use wall-clock expiry polling to survive system sleep/wake

novomanish:
## Summary

- Fixes the Copilot token refresh loop failing to recover after system sleep (laptop sleep/wake or Docker host sleep)
- The previous loop used `setTimeout` with `refresh_in` seconds, which relies on the **monotonic clock**. On Linux/Docker, the monotonic clock does **not** advance during system sleep — so the timer would never fire after a long sleep, leaving the token expired until the container restarted
- Switch to **wall-clock polling**: track `refreshAtMs = expires_at * 1000 - 60s`, sleep in 30s chunks (`REFRESH_POLL_INTERVAL_MS`), and recheck `Date.now()` on each wake. Within 30s of wakeup, the loop detects `msUntilRefresh <= 0` and refreshes immediately

## Test plan

- [ ] Start the server, sleep laptop for >1h, wake — token should refresh within ~30s without restarting the container
- [ ] Verify normal proactive refresh still fires ~60s before token expiry
- [ ] Verify failed refresh retries after 15s as before

### DIFF
--- a/src/lib/token.ts
+++ b/src/lib/token.ts
@@ -44,7 +44,7 @@ export const setupCopilotToken = async () => {
     return
   }
 
-  const { token, refresh_in } = await getCopilotToken()
+  const { token, expires_at } = await getCopilotToken()
   state.copilotToken = token
 
   // Display the Copilot token to the screen
@@ -58,7 +58,7 @@ export const setupCopilotToken = async () => {
   const controller = new AbortController()
   copilotRefreshLoopController = controller
 
-  runCopilotRefreshLoop(refresh_in, controller.signal)
+  runCopilotRefreshLoop(expires_at, controller.signal)
     .catch(() => {
       consola.warn("Copilot token refresh loop stopped")
     })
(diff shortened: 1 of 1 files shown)


documanish:
@caozhiyuan There is a real problem when computer sleeps, I have to wait a long time before the loop refreshes the token. I think we can update the the refresh loop using actual clock instead of process timer.

caozhiyuan:
When the computer goes to sleep, the application is also paused. When it resumes from sleep (after a long period of sleep), the application refreshes the token. If the network connection is slow, the first token acquisition will fail. However, after 15 seconds, it will retry and succeed this time. So what is your problem? @novomanish

caozhiyuan:
@documanish

documanish:
> When the computer goes to sleep, the application is also paused. When it resumes from sleep (after a long period of sleep), the application refreshes the token. If the network connection is slow, the first token acquisition will fail. However, after 15 seconds, it will retry and succeed this time. So what is your problem? @novomanish

**Worst case wait: ~29 minutes NOT 15s**
`let nextRefreshDelayMs = (refreshIn - 60) * 1000 = 29 minutes`
*assuming typical refresh_in= 1800 (30 minutes)

The app sets a single timer to wake up and refresh the token, roughly 29 minutes after the last refresh. While the computer is asleep, that timer is effectively frozen — it doesn't count down. So when you wake the computer, the app doesn't know any time passed and just picks up where it left off, with most of those 29 minutes still on the clock.

If the token expired while the computer was asleep, the app keeps using the stale token, causing 401 errors. It won't try to refresh until that 29-minute timer finally runs out — which, depending on when you fell asleep, could mean waiting close to the full 29 minutes after waking up.

Only once the timer fires does the app attempt a network call to get a new token. If the network is slow at that point, it fails and waits another 15 seconds before retrying. The 15-second retry is real, but it only becomes relevant after the long wait is already over.

caozhiyuan:
@documanish ok  , i will review it

PR #175, opened 2026-04-11
Strip context_management from Messages API payload

gzenz:
## Summary

- Strip `context_management` field from Messages API payloads before forwarding to the Copilot Messages endpoint
- The upstream Copilot API rejects this field with "Extra inputs are not permitted"
- Claude Code started sending `context_management` recently; this fix prevents 400 errors

## Context

`context_management` is a valid Anthropic API field but is not supported by the Copilot Messages endpoint. Since `AnthropicMessagesPayload` doesn't declare it, TypeScript doesn't catch it at compile time, but the field passes through `JSON.stringify(payload)` at runtime.

The fix adds a `stripUnsupportedPayloadFields` function in `preprocess.ts` that removes unsupported top-level fields before the request is sent. This is extensible for future fields that may need stripping.

## Test plan

- [ ] Verify Claude Code sessions work without "Extra inputs are not permitted" errors
- [ ] Verify no regression in existing message preprocessing (thinking blocks, cache control, tool results)
- [ ] Run `npm run lint` and `npm run build` to confirm no type/lint errors

### DIFF
--- a/src/routes/messages/preprocess.ts
+++ b/src/routes/messages/preprocess.ts
@@ -218,11 +218,27 @@ const filterAssistantThinkingBlocks = (
   }
 }
 
+// Strip fields from the top-level payload that the Copilot Messages API
+// rejects with "Extra inputs are not permitted". These are valid Anthropic API
+// fields but not (yet) supported by the Copilot Messages endpoint.
+const stripUnsupportedPayloadFields = (
+  payload: AnthropicMessagesPayload,
+): void => {
+  // Valid Anthropic API fields not yet supported by the Copilot Messages endpoint.
+  // They arrive at runtime via JSON pass-through despite not being in our type.
+  for (const key of ["context_management"]) {
+    if (key in payload) {
+      Reflect.deleteProperty(payload, key)
+    }
+  }
+}
+
 export const prepareMessagesApiPayload = (
   payload: AnthropicMessagesPayload,
   selectedModel?: Model,
 ): void => {
   stripCacheControl(payload)
+  stripUnsupportedPayloadFields(payload)
   filterAssistantThinkingBlocks(payload)
 
   // https://platform.claude.com/docs/en/build-with-claude/extended-thinking#extended-thinking-with-tool-use


caozhiyuan:
Copilot supports context management. Please share the error you encountered, along with the request body. @gzenz

gzenz:
Thanks for the quick response! If Copilot now supports `context_management`, that's great news -- the error I was seeing was `"Extra inputs are not permitted"` when Claude Code started sending `context_management` in the request body.

Given that Copilot supports it now, this PR may not be needed. I've updated my local preload to use a retry-on-failure approach instead: it sends the full body first and only strips unsupported fields if the proxy rejects them. That way `context_management` passes through when supported.

I'll close this PR if you confirm there's nothing to fix on the copilot-api side. Thanks!

caozhiyuan:
@gzenz in github copilot extension, support open messages api context edit. and i test context management works ok on latest claude code version.

gzenz:
@caozhiyuan puzzling - i had issues with this a few weeks back, and since then am using this work around - so i thought i finally open a PR for it :) But I just removed it and added error logging. I'm now doing a larger refactor and yes - I see no errors at all. Now I haven't kept logs from a few weeks back, but I think it happened when a lot of sub agents were running. Maybe it was just a flake? Or it's fixed by now? Curious. I'll keep monitoring it for a few days and then report back, or close this.

gzenz:
@caozhiyuan ok you're right - it works flawlessly. Sorry for causing all the turmoil :) closing.

newtee0517:
@gzenz i get this error {"error":{"message":"{\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"context_management: Extra inputs are not
     permitted\"},\"request_id\":\"req_vrtx_011Ca75jC\"}","type":"error"}}  , sees route to vrtx will raise error , if route to bedrock ，it‘s ok

gzenz:
I started getting (with claude code 2.1.110): API Error: 400 {"error":{"message":"{\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"context_hint: Extra
     inputs are not permitted\"},\"request_id\":\"req_vrtx_011.....\"}","type":"error"}}
     
so this whole thing seems to be somewhat flaky, still.

newtee0517:
@gzenz old claude code 2.1.98 version is ok . Too strange

PR #200, opened 2026-04-20
feat: show model name badge in incoming request log

go9sky:
## What

Replace the default Hono `logger()` middleware with a custom `modelAwareLogger` that prints the requested model name as a colored badge on the incoming `<--` log line.

## Why

When proxying multiple model requests, it's useful to see at a glance which model is being requested without digging into logs.

## How

- Clone the request body via `Request.clone()` — zero impact on  the actual handler, the original body is untouched.
- Use `getColorEnabled()` from Hono to respect the terminal's  color support; degrades gracefully to plain `[model-name]`.

## Example

Before:
```
  <-- POST /v1/messages?beta=true
  --> POST /v1/messages?beta=true 200 5s
```

After:
```
  <-- POST /v1/messages?beta=true  [claude-opus-4-5]
  --> POST /v1/messages?beta=true 200 5s
```

On the terminal, like this：
image
image


caozhiyuan:
@go9sky Intercepting the request stream is not a good approach

go9sky:
I added this tag to clearly know which model `Claude code` or other Agents is actually calling. Of course, you can add a configuration or parameter to automatically disable this tag.

caozhiyuan:
The problem is that you intercepted the request stream globally and copied and attempted to parse the stream. @go9sky

caozhiyuan:
You can print logs in handler.ts. @go9sky

go9sky:
> You can print logs in handler.ts. @go9sky

You are right. I've reworked the implementation based on your suggestions:
- **Removed the global stream interception** — no more `Request.clone()` in middleware
- **Model is logged from within each handler** (`messages`, `chat-completions`, `responses`, `provider/messages`) after the payload is already parsed, so there's zero extra overhead
- The middleware now uses `AsyncLocalStorage` to defer printing the `<-- METHOD PATH` line until the handler calls `logRequestModel(model)`, so the model name appears on the same log line. For routes without a model (e.g. `GET /models`), it falls back to printing the line without a badge after the handler returns.

caozhiyuan (changes requested):
I like the goal here, but I see two blocking issues in the current implementation.

1. `modelAwareLogger` changes Hono's access-log semantics. Hono's built-in logger emits `<-- METHOD PATH` before `await next()`, but this implementation defers that line until `logRequestModel()` is called or until after `await next()` completes. That means routes without `logRequestModel()` now effectively log `handler -> <-- -> -->`, `/v1/messages` can delay the incoming log until after manual approval, and early uncaught failures can lose the incoming log entirely. I would preserve Hono's original timing, store the resolved model in AsyncLocalStorage, and append the badge to the outgoing line (or emit a second model line once resolved) instead of delaying the incoming access log.

2. The badge writes raw user-controlled `payload.model` values into the terminal log. Newlines and ANSI control sequences can forge or corrupt access logs. Please sanitize or escape control characters and cap the logged model length before formatting the badge.

Suggested direction:
- restore immediate `<-- METHOD PATH` logging in the middleware
- change `logRequestModel()` into a `setRequestModel()` helper that only stores a sanitized model in AsyncLocalStorage
- append the badge to the `-->` line, or log the model on a separate line once it is known
- add regression tests for non-model routes and model sanitization

@go9sky

caozhiyuan:
Just use consola info directly in handler.ts ; if you want a specific style, design it yourself. @go9sky

caozhiyuan:
@go9sky feat add in 1.6.6

go9sky:
> @go9sky feat add in 1.6.6

Thanks for the update and for implementing this request. I’ve seen your changes and understood the points about log timing and sanitization.

PR #209, opened 2026-04-22
fix: skip [DONE] terminator in messages API stream

bdeb1337:
When streaming through the `/v1/messages` endpoint (`handleWithMessagesApi`), the backend appends a `data: [DONE]` terminator after `message_stop`: an openAI streaming convention.

added skipping `[DONE]` and empty `data` lines in the messages API flow, matching the existing behavior in `handleWithChatCompletions` which already handled this.


caozhiyuan:
@bdeb1337 What's the problem now?

PR #219, opened 2026-04-27
增加桌面端支持

fangzc1:
代码解耦，无侵入
image
image


caozhiyuan:
@fangzc1 Can the plan or spec be deleted? Use AI to translate non-English parts elsewhere.

PR #222, opened 2026-04-29
feat: desktop

fangzc1:
image


fangzc1:
已修复 bun run typecheck 问题

caozhiyuan:
@fangzc1 There is also the issue of Chinese characters in the code; can we use i18n or English instead? Can auth.ts reuse the original process, and does opencode auth support it?

caozhiyuan:
Or you can PR to the electron branch, I'll make changes when I have time  @fangzc1

fangzc1:
Although both files implement the GitHub OAuth Device Flow, their abstraction levels differ significantly:

- `src/auth.ts` serves as the entry point for the command, while the core logic resides within `src/lib/token.ts`.
- In contrast, `desktop/electron/auth.ts` consolidates the entire OAuth process within a single file, resulting in a cleaner structure better suited for Electron IPC (Inter-Process Communication) calls.
- Forcibly reusing `src/lib/token.ts` would necessitate pulling in the server-side dependency chain (including state, paths, etc.), which would introduce unnecessary overhead and ultimately prove counterproductive.
@caozhiyuan

fangzc1:
I will complete the i18n implementation.

PR #223, opened 2026-04-29
feat: desktop

fangzc1:
i18n: Supported
opencode: Not supported
Although both files implement the GitHub OAuth Device Flow, their abstraction levels differ significantly:

- `src/auth.ts` serves as the entry point for the command, while the core logic resides within `src/lib/token.ts`.
- In contrast, `desktop/electron/auth.ts` consolidates the entire OAuth process within a single file, resulting in a cleaner structure better suited for Electron IPC (Inter-Process Communication) calls.
- Forcibly reusing `src/lib/token.ts` would necessitate pulling in the server-side dependency chain (including state, paths, etc.), which would introduce unnecessary overhead and ultimately prove counterproductive.


caozhiyuan:
@fangzc1 pr to https://github.com/caozhiyuan/copilot-api/tree/electron

caozhiyuan:
There are quite a few issues; I've synchronized them to this branch: https://github.com/caozhiyuan/copilot-api/commits/electron/. It still needs a lot of refinement, such as the menu bar being displayed on Windows and ANSI log encoding issues. @fangzc1

fangzc1:
You're right, I learn from you

PR #247, opened 2026-05-19
fix: Fix missing object attribute in openai /chat/completions

diogosilva30:
## Description
Fix missing `"object": "chat.completion"` field in `/v1/chat/completions` endpoint response.

## Problem
The endpoint returns JSON without the required `"object": "chat.completion"` field, causing validation errors in OpenAI-compatible clients (Pydantic-AI, LangChain, etc.).

## Solution
Explicitly add `"object": "chat.completion"` to the response before returning it to ensure full OpenAI API specification compliance.

## Testing
Clients like Pydantic-AI no longer fail with:

```shell
Input should be 'chat.completion' [type=literal_error, input_value=None, input_type=NoneType]
```

### DIFF
--- a/src/routes/chat-completions/handler.ts
+++ b/src/routes/chat-completions/handler.ts
@@ -74,7 +74,12 @@ export async function handleCompletion(c: Context) {
   if (isNonStreaming(response)) {
     debugJson(logger, "Non-streaming response:", response)
     recordUsage(normalizeOpenAIUsage(response.usage))
-    return c.json(response)
+    // Ensure the response includes the required "object" field for OpenAI compatibility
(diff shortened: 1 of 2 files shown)


diogosilva30:
@caozhiyuan please take a look and review

caozhiyuan:
@diogosilva30 Is it necessary to add the stream event?

caozhiyuan:
@diogosilva30 Use the message API or responses API; the chat completions interface in Copilot is not a standard protocol and contains a thinking signature. If this signature is missing, the model's intelligence will degrade.  Moreover, the chat completions interface has already been deprecated by OpenAI.

PR #257, opened 2026-05-29
fix(responses): guard input images against 413 payload errors

Marvae:
## Summary

Prevent `copilot-api` `/v1/responses` requests from getting stuck in repeated `413 Payload Too Large` failures when inline base64 `input_image` data URLs make the request body too large.

This PR guards the Copilot Responses proxy path in two places:

- before forwarding, replace inline images that exceed the model's `max_prompt_image_size` with a tiny valid placeholder image
- after an upstream 413, replace the remaining inline images with the same placeholder and retry once

## Problem

The failure this PR targets is a 413 payload-size loop on the `copilot-api` `/v1/responses` route. When a client sends inline `input_image` data URLs through `POST /v1/responses`, GitHub Copilot can reject the request with:

```text
unexpected status 413 Payload Too Large:
{"error":{"message":"failed to parse request","code":""}},
url: http://localhost:4141/v1/responses
```

One concrete way to hit this is Codex usage: images from tools such as `view_image` are stored in the thread history as inline base64 `input_image` data URLs. Once a large image is in history, every retry/resume sends the same oversized payload again. Codex retries 5 times and fails each time. Manual compaction may not recover the thread either, because the compaction request still has to send the same image-containing history before a compacted summary can be produced.

In practice this can make the affected session unusable until the inline image data is removed from the payload.

## Investigation

Copilot model metadata exposes vision limits under `capabilities.limits.vision`, including:

```ts
vision?: {
  max_prompt_image_size?: number
  max_prompt_images?: number
  supported_media_types?: string[]
}
```

For affected models used in this flow, `max_prompt_image_size` was `3145728` bytes.

Local payload checks showed:

- a single valid inline image below the per-image limit was accepted
- a valid image above the per-image limit reproduced the 413 behavior
- multiple tiny images were accepted, so image count alone was not the direct trigger
- larger total request bodies could trigger 413 even when individual images were valid

The most important failure mode was historical `view_image` tool output being replayed as `function_call_output.output[].input_image` data URLs. Those inline image payloads are expensive in the JSON request body and can be resent on every retry/resume.

## Fix

This PR adds two defensive checks in the `/v1/responses` route:

1. Before forwarding to Copilot, replace inline `input_image` data URLs whose decoded bytes exceed the model's `max_prompt_image_size`.
2. If Copilot still rejects the request with HTTP 413, replace the remaining inline `input_image` data URLs and retry once.

The replacement keeps `type: "input_image"` and `detail: "low"`, but swaps the original large `image_url` for a tiny valid PNG placeholder that reads `Image too large / Redacted`.

The size check estimates decoded bytes from the data URL string length and base64 padding, without decoding or capturing the full base64 payload.

## Local Validation

This was first patched into a local `copilot-api` install and used against an affected client session that was repeatedly failing with 413.

Before the patch, the session retried the same oversized payload until the turn failed:

```text
Reconnecting... 1/5 (unexpected status 413 Payload Too Large ...)
...
Reconnecting... 5/5 (unexpected status 413 Payload Too Large ...)
turn.failed
```

After the patch, the same session resumed successfully and continued producing responses. The local proxy logged image omission before forwarding the recovered request:

```text
Omitted 8 input image(s) after Copilot Responses rejected the payload as too large
```

The source changes are covered by focused unit tests for:

- replacing images above `max_prompt_image_size`
- preserving multiple small images before any 413
- retrying once with readable placeholder images after a 413 response
- estimating base64 decoded bytes from data URL length and padding

I also tested the readable PNG placeholder against the real Copilot `/responses` endpoint and it was accepted.

## Reference

The Copilot SDK image-input documentation describes richer runtime behavior: images that exceed model constraints may be resized while preserving aspect ratio, quality-reduced, or skipped if they still cannot be brought within limits. It also documents `capabilities.limits.vision.max_prompt_images` and `max_prompt_image_size` as model vision limits:

- https://docs.github.com/en/copilot/how-tos/copilot-sdk/image-input
- https://github.com/github/copilot-sdk/blob/eb3c54c3a6871a5a654ae230555c042501b13018/nodejs/src/types.ts#L2323-L2327

This PR takes the smaller dependency-free approach of replacing inline image data URLs when they cannot be safely forwarded. A possible follow-up would be to discuss whether `copilot-api` should also attempt image compression, resizing, or downsampling before replacement. That could preserve more visual context, but it would also add image-processing behavior to the proxy.

### DIFF
--- a/src/routes/responses/handler.ts
+++ b/src/routes/responses/handler.ts
@@ -21,6 +21,7 @@ import {
   type ResponsesResult,
   type ResponseStreamEvent,
 } from "~/services/copilot/create-responses"
+import { HTTPError } from "~/lib/error"
 
 import { createStreamIdTracker, fixStreamIds } from "./stream-id-sync"
 import {
@@ -28,6 +29,8 @@ import {
   compactInputByLatestCompaction,
   getResponsesTransportForModel,
   getResponsesRequestOptions,
+  sanitizeAllInputImages,
+  sanitizeOversizedInputImages,
 } from "./utils"
 
 const logger = createHandlerLogger("responses-handler")
@@ -91,6 +94,16 @@ export const handleResponses = async (c: Context) => {
     )
   }
 
+  const sanitizedImageCount = sanitizeOversizedInputImages(
+    payload,
+    selectedModel?.capabilities.limits.vision?.max_prompt_image_size,
+  )
+  if (sanitizedImageCount > 0) {
+    logger.warn(
+      `Omitted ${sanitizedImageCount} oversized input image(s) before forwarding to Copilot Responses`,
+    )
+  }
+
   applyResponsesApiContextManagement(
     payload,
     selectedModel?.capabilities.limits.max_prompt_tokens,
@@ -104,13 +117,41 @@ export const handleResponses = async (c: Context) => {
     await awaitApproval()
   }
 
-  const response = await responsesHandlerDependencies.createResponses(payload, {
+  const responseOptions = {
     vision,
     initiator,
     requestId,
     sessionId: sessionId,
     transport: responsesTransport,
-  })
+  }
+  let response: Awaited<ReturnType<typeof createCopilotResponses>>
+  try {
+    response = await responsesHandlerDependencies.createResponses(
+      payload,
+      responseOptions,
+    )
+  } catch (error) {
+    if (!(error instanceof HTTPError) || error.response.status !== 413) {
+      throw error
+    }
+
+    const retrySanitizedImageCount = sanitizeAllInputImages(payload)
+    if (retrySanitizedImageCount === 0) {
+      throw error
+    }
+
+    logger.warn(
+      `Omitted ${retrySanitizedImageCount} input image(s) after Copilot Responses rejected the payload as too large`,
+    )
+    const retryOptions = {
+      ...responseOptions,
+      vision: getResponsesRequestOptions(payload).vision,
+    }
+    response = await responsesHandlerDependencies.createResponses(
+      payload,
+      retryOptions,
+    )
+  }
 
   if (isStreamingRequested(payload) && isAsyncIterable(response)) {
     logger.debug("Forwarding native Responses stream")
--- a/src/routes/responses/utils.ts
+++ b/src/routes/responses/utils.ts
@@ -80,6 +80,26 @@ export const hasVisionInput = (payload: ResponsesPayload): boolean => {
 const DATA_URL_PREFIX = "data:"
 const BASE64_MARKER = ";base64,"
 const IMAGE_MEDIA_TYPE_PATTERN = /^image\/[a-zA-Z0-9.+-]+$/
+// Static 96x32 PNG reading "Image too large / Redacted".
+const REDACTED_IMAGE_PLACEHOLDER_DATA_URL =
+  "data:image/png;base64,"
+  + [
+    "iVBORw0KGgoAAAANSUhEUgAAAGAAAAAgCAMAAADaHo1mAAADAFBMVEX///8fKTfR1dsAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+    "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+    "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+    "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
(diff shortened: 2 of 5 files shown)


caozhiyuan:
@Marvae Accurate calculation is resource-intensive; can we estimate directly using the data string? Also, replacing the original image type is cumbersome; can we generate a base64 placeholder? Mark the image as too large to read.

Marvae:
@caozhiyuan Updated the size check to use the data URL string length + base64 padding, without decoding/capturing the full base64 payload.

For the placeholder, I first tried preserving the original MIME type with a tiny synthetic payload, but Copilot `/responses` returns 400 if the bytes are not a valid image. So I used a tiny valid PNG placeholder instead, while keeping `type: "input_image"` and `detail: "low"`.

One concern with this placeholder approach is that the model can no longer read why the image was replaced unless the placeholder itself has readable text. Do you think the tiny valid placeholder is enough, or should the placeholder image include text like "Image too large / Redacted"?

caozhiyuan:
should the placeholder image include text like "Image too large / Redacted"? yes.
and don't catch , in next round will error again .  you can prehandle it before call v1/responses @Marvae

Marvae:
@caozhiyuan For "don't catch", do you mean we should also pre-handle the total request body size before calling Copilot `/responses`?

The model metadata gives `max_prompt_image_size`, so single-image oversized cases are already handled before forwarding. I have not found a model-provided total body-size limit, so the current 413 retry is only for that total-payload case.

caozhiyuan:
@Marvae What I mean is that during the second round of requests, if the image has not been processed, it may lead to a cache hit anomaly or could be rejected in the second round.

Marvae:
@caozhiyuan I think the 413 catch is covering a different case.

Preflight already handles images over `max_prompt_image_size` before the first Copilot call. The retry path is for the case where each image is below `max_prompt_image_size`, but the total JSON body is still too large and Copilot returns 413.

I tested this with the current branch against real Copilot `/responses`: two images were each below the per-image limit, but the request body was about 5.8MB. The first upstream call failed, the handler replaced the remaining inline images in the same payload, and the second upstream call returned 200.

This is not just a theoretical fallback. I hit both cases in real Codex usage:

- one inline image was larger than `max_prompt_image_size`, so preflight handles it
- later, individual images were below that limit, but the full request body was still too large and Copilot returned 413, so the retry path handled it

So the retry request does not carry the original inline images. I kept this as a fallback because I have not found a model-provided total body-size limit to pre-handle this case safely.

Marvae:
@caozhiyuan Thanks for the review!

caozhiyuan:
@Marvae released

Marvae:
@caozhiyuan I dug deeper and tested this through the official Copilot CLI as well.

With `gpt-5.5`, `gpt-5.4`, `gpt-5.2-codex`, and `gpt-5.3-codex`, two valid PNG attachments were each below `max_prompt_image_size`, but the request failed with:

```text
The attached image or document is too large to send natively through CAPI Responses.
Try a smaller attachment or fewer attachments.
(6.3 MB request; 5.0 MB limit)
```

So `max_prompt_image_size` seems to be per-image, while CAPI Responses also has a 5MB native request body limit.

This looks like a separate total-payload guard from the per-image check. Do you think it should be handled as a follow-up?

<details>
<summary>CLI command and output</summary>

```sh
# generated two valid PNG files first
# a.png 2434189
# b.png 2433855

COPILOT_HOME=/tmp/copilot-cli-proof-one/home \
copilot -p 'Reply with only OK.' \
  --model gpt-5.5 \
  --attachment ./a.png \
  --attachment ./b.png \
  --allow-all-tools \
  --no-custom-instructions \
  --no-auto-update \
  --log-level debug \
  --log-dir ./logs \
  --silent
```

```text
exit: 1

Execution failed: Error: Failed to get response from the AI model; retried 5 times (total retry wait time: 5.73 seconds) Last error: Unknown error

... Using model: gpt-5.5
... "max_prompt_image_size": 3145728,
... "max_prompt_images": 1,
... Checking if image is valid size and dimensions. Max size: 3145728, Max dimension: 2000
... Image is within size and dimension limits.
... Checking if image is valid size and dimensions. Max size: 3145728, Max dimension: 2000
... Image is within size and dimension limits.
... Making responses request with input: [
... StreamingErrorRetryProcessor: Non-API error detected (retry 1): The attached image or document is too large to send natively through CAPI Responses. Try a smaller attachment or fewer attachments. (6.3 MB request; 5.0 MB limit)
... Making responses request with input: [
... StreamingErrorRetryProcessor: Non-API error detected (retry 2): The attached image or document is too large to send natively through CAPI Responses. Try a smaller attachment or fewer attachments. (6.3 MB request; 5.0 MB limit)
... Making responses request with input: [
... StreamingErrorRetryProcessor: Non-API error detected (retry 3): The attached image or document is too large to send natively through CAPI Responses. Try a smaller attachment or fewer attachments. (6.3 MB request; 5.0 MB limit)
... Making responses request with input: [
... StreamingErrorRetryProcessor: Non-API error detected (retry 4): The attached image or document is too large to send natively through CAPI Responses. Try a smaller attachment or fewer attachments. (6.3 MB request; 5.0 MB limit)
... Making responses request with input: [
... StreamingErrorRetryProcessor: Non-API error detected (retry 5): The attached image or document is too large to send natively through CAPI Responses. Try a smaller attachment or fewer attachments. (6.3 MB request; 5.0 MB limit)
... Making responses request with input: [
... StreamingErrorRetryProcessor: Non-API error detected (retry 6): The attached image or document is too large to send natively through CAPI Responses. Try a smaller attachment or fewer attachments. (6.3 MB request; 5.0 MB limit)
```

</details>

caozhiyuan:
What I mean is that failing to maintain consistent prefixes will negatively impact cache hit rates. @Marvae Therefore, it is processed before each request.

PR #260, opened 2026-05-30
fix(messages): strip eager_input_streaming and cache_control to fix 400s

bdeb1337:
This PR fixes two separate 400s hitting Haiku 4.5, Sonnet 4.5, and Sonnet 4.6 when using `anthropic` provider w/Zed through the proxy:

- ```
  tools.N.custom.eager_input_streaming: Extra inputs are not permitted
  ```

Zed injects `eager_input_streaming: true` into every tool definition and Copilot's
backend doesn't accept it. Zed's own `copilot_chat` provider, opencode, kilocode, and a few other proxies all strip it.

- ```
  cache_control: Extra inputs are not permitted
  ```

The existing `stripCacheControl` only trimmed `scope` off system blocks. Haiku rejects
`cache_control` anywhere in the payload: tools, message content, `tool_result`
inner blocks, top-level.

---

### Changes: 

- **`stripToolEagerInputStreaming`** (new): strips `eager_input_streaming` from all tools
- **`stripCacheControl`** (expanded): now removes `cache_control` from the full payload tree
- Added `eager_input_streaming?: boolean` to `AnthropicTool` for type safety

### DIFF
--- a/src/routes/messages/anthropic-types.ts
+++ b/src/routes/messages/anthropic-types.ts
@@ -132,6 +132,7 @@ export interface AnthropicTool {
   description?: string
   input_schema: Record<string, unknown>
   defer_loading?: boolean
+  eager_input_streaming?: boolean
   cache_control?: AnthropicCacheControl | null
 }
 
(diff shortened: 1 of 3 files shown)


caozhiyuan:
@bdeb1337 The removal of Cache-Control results in context cache misses.

bdeb1337:
Thank you so much for checking this so quickly!
You are correct, sorry for that.. it's not the entire cache-control that should be removed/stripped but only the `scope` inside of it. I'll change the fix and try a new PR.

PR #261, opened 2026-05-31
fix(messages): strip eager_input_streaming and cache_control scope

bdeb1337:
This PR fixes two separate 400s hitting Haiku 4.5, Sonnet 4.5, and Sonnet 4.6 when using `anthropic` provider w/Zed through the proxy:

- ```
  tools.N.custom.eager_input_streaming: Extra inputs are not permitted
  ```

Zed injects `eager_input_streaming: true` into every tool definition and Copilot's
backend doesn't accept it. Zed's own `copilot_chat` provider, opencode, kilocode, and a few other proxies all strip it.

- ```
  cache_control: Extra inputs are not permitted
  ```

instead of stripping the entire cache_control block like my other PR, this just extends the existing scope-strip to tools and message content blocks, not only system blocks.

## Fix

- **`stripToolEagerInputStreaming`** (new) — strips `eager_input_streaming` from all tools
- **`stripCacheControlScope`** (expanded from `stripCacheControl`) — strips `scope` from
  `cache_control` across system blocks, tools, and message content
- Added `eager_input_streaming?: boolean` to `AnthropicTool` for type safety

thank you in advance for your time and for reviewing this.

### DIFF
--- a/src/routes/messages/anthropic-types.ts
+++ b/src/routes/messages/anthropic-types.ts
@@ -132,6 +132,7 @@ export interface AnthropicTool {
   description?: string
   input_schema: Record<string, unknown>
   defer_loading?: boolean
+  eager_input_streaming?: boolean
   cache_control?: AnthropicCacheControl | null
 }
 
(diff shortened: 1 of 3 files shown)


caozhiyuan:
@bdeb1337 Please send the request message that caused the exception, start with the -v parameter, and the logs are located at .local/share/copilot-api/logs under user

bdeb1337:
Thank you for your response. I will try to split the two issues into separate PR's and provide proper evidence with the `-v` parameter to reproduce each one first. I have to dig a bit deeper because I am making too many assumptions here. Please excuse me as I am new to these internal workings.

PR #263, opened 2026-05-31
fix: strip top-level cache_control for unsupported models and scope from all blocks

bdeb1337:
Hello @caozhiyuan,
I took some time to try and pinpoint the actual cache_control problems some more.

Here are two reproducible curl commands to see the issues at hand:

```bash
# 1. Top-level cache_control → 400 on sonnet-4.6 (unpatched)
curl -s -w "\n%{http_code}\n" -X POST http://localhost:4141/v1/messages \
  -H "Content-Type: application/json" \
  -d '{
    "model": "claude-sonnet-4.6",
    "max_tokens": 32,
    "stream": false,
    "cache_control": {"type": "ephemeral"},
    "system": [{"type": "text", "text": "Reply with OK.", "cache_control": {"type": "ephemeral", "ttl": "1h"}}],
    "tools": [
      {
        "name": "get_weather",
        "description": "Get weather",
        "input_schema": {"type": "object", "properties": {"city": {"type": "string"}}},
        "cache_control": {"type": "ephemeral", "ttl": "1h"}
      }
    ],
    "messages": [{"role": "user", "content": "Say OK"}]
  }'

# 2. scope on tool cache_control → 400 (unpatched, scope not stripped on tools)
curl -s -w "\n%{http_code}\n" -X POST http://localhost:4141/v1/messages \
  -H "Content-Type: application/json" \
  -d '{
    "model": "claude-opus-4.7",
    "max_tokens": 32,
    "stream": false,
    "tools": [
      {
        "name": "get_weather",
        "description": "Get weather",
        "input_schema": {"type": "object", "properties": {"city": {"type": "string"}}},
        "cache_control": {"type": "ephemeral", "ttl": "1h", "scope": "session"}
      }
    ],
    "messages": [{"role": "user", "content": "Say OK"}]
  }'
```

These both return `HTTP error: { message: 'cache_control: Extra inputs are not permitted' }`.

Regarding `top-level cache_control`: I have noticed only Opus 4.6+ can handle these request types reliably so now I would only strip this for unsupported models.

Regarding `scope on tool cache_control`: this isn't supported on any model via github copilot and should be stripped 

Both these curls work against this PR and provide compatibility for this matter using Zed with the tools' native Anthropic provider.

The other issue remaining is regarding `eager_input_streaming` but I will try to address that in another PR.

Sorry if I still missed something here: I'm open for discussion/suggestions.

Kind regards.

### DIFF
--- a/src/routes/messages/preprocess.ts
+++ b/src/routes/messages/preprocess.ts
@@ -730,18 +730,61 @@ const hasToolRef = (block: AnthropicToolResultBlock) => {
   )
 }
 
-// Strip cache_control from system content blocks as the
-// Copilot Messages API does not support them (rejects extra fields like scope).
-// commit by nicktogo
+// Models whose Copilot Messages API endpoints reject top-level cache_control.
+// Pattern: only Opus 4.6+ reliably accepts top-level cache_control.
+const TOP_LEVEL_CACHE_CONTROL_UNSUPPORTED_MODELS = [
+  "claude-sonnet-4.5",
+  "claude-sonnet-4.6",
+  "claude-opus-4.5",
+  "claude-haiku-4.5",
+]
+
+// Strip unsupported cache_control fields for Copilot Messages API compatibility.
+// - `scope` in block-level cache_control is always rejected and should be stripped everywhere.
+// - Top-level `cache_control` is rejected by some models and should be stripped.
 const stripCacheControl = (payload: AnthropicMessagesPayload): void => {
+  if (
+    payload.cache_control
+    && TOP_LEVEL_CACHE_CONTROL_UNSUPPORTED_MODELS.includes(payload.model)
+  ) {
+    delete payload.cache_control
+  }
   if (Array.isArray(payload.system)) {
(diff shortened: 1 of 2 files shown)


caozhiyuan:
@bdeb1337 You need to give me the actual request Zed transmits; those are standard parameters. The scope is what claude code used to transmit, and I don't believe Zed would transmit scope.

caozhiyuan:
https://github.com/zed-industries/zed/blob/main/crates/anthropic/src/anthropic.rs#L525

caozhiyuan:
if top level cache_control be removed , need add 4 cache_control to messages and system block.

caozhiyuan:
You can raise the issue as an issue ticket. If you want to fix it, please do so after I confirm the problem, rather than fixing it blindly. For now, it will be closed. @bdeb1337

bdeb1337:
Agreed. I will create an issue with proper logs and start over. I will only work on implementation after you confirm the problem fully. Please excuse me for the inconvenience.

PR #265, opened 2026-06-01
feat: write env file on start with optional shell rc sourcing

Vivswan:
## Summary
- Generate a dotenv file (`copilot-api.env`) and a POSIX wrapper (`copilot-api.env.sh`) in the app dir on server start, both written with `0600` permissions so secrets aren't world-readable.
- Add `--source-env` flag to append a managed, idempotent source block to the user's shell rc (`.zshrc`/`.bash_profile`/`.profile`).
- Add `--claude-model`, `--claude-small-model`, and `--gpt-model` flags to control the model IDs written to the env file.
- Refactor `--claude-code` clipboard command generation to share `buildClaudeEnvVars`.

## Test plan
- [x] `bun test tests/env-file.test.ts` passes
- [x] `bun run start --source-env` writes the env files and appends the rc block once (idempotent)
- [x] Generated dotenv values are single-quoted/escaped to stay injection-safe


caozhiyuan:
@Vivswan This approach isn't very user-friendly, right? Usually, configurations are set in settings.json.

Vivswan:
The main goal is to ensure that Claude and Codex are automatically ready to use once the copilot-api proxy starts. Do you have any suggestions on how we can accomplish that?

caozhiyuan:
@Vivswan What if the model is switched later? Generally, model configurations are set in claude code settings.json or codex config.toml.

PR #271, opened 2026-06-03
Add configurable model metadata overrides

Vivswan:
## Summary
- add config/admin API support for per-model metadata overrides
- deep-merge overrides into cloned model metadata while preserving model ids
- apply overrides in model listing and model resolution paths, including Codex provider models

## Tests
- bun run typecheck
- bun run lint:all
- bun test

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -4,6 +4,8 @@ import fs from "node:fs"
 
 import { PATHS } from "./paths"
 
+export type ModelMetadataOverride = Record<string, unknown>
+
 export interface AppConfig {
   auth?: {
     apiKeys?: Array<string>
@@ -11,6 +13,7 @@ export interface AppConfig {
   }
   providers?: Record<string, ProviderConfig>
   modelMappings?: Record<string, string>
+  modelOverrides?: Record<string, ModelMetadataOverride>
   extraPrompts?: Record<string, string>
   smallModel?: string
   useResponsesApiContextManagement?: boolean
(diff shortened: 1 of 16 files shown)


caozhiyuan:
What is the purpose of this feature @Vivswan

Vivswan:
I want to ability to edit the max input and output token parameters and some other parameters

caozhiyuan:
Which AI programming tool are you using? It seems there's nowhere that reads those parameters. @Vivswan

Vivswan:
some Copilot models support 1M context, but compatibility clients may still see 200K metadata.

image


Example: Claude Code selects `claude-opus-4.6-1m`, but `/context` shows `31.8k/200k`. With this PR I can fix the advertised metadata:

```json
{
  "modelOverrides": {
    "claude-opus-4.6-1m": {
      "capabilities": {
        "limits": {
          "max_context_window_tokens": 1000000,
          "max_prompt_tokens": 1000000,
          "max_output_tokens": 32000
        }
      }
    }
  }
}
```

caozhiyuan:
@Vivswan Claude Code does not read the configuration here. You need to enable 1M context and set the compaction window size through configuration in Claude Code.

caozhiyuan:
Or you can configure the model ID like this: claude-opus-4.6[1m], just add [1m] at the end of the model name @Vivswan

Vivswan:
The original concept was working for me.
But yes, using claude-opus-4.6[1m] is simpler. I will try that

PR #272, opened 2026-06-03
Log requested model for messages

Vivswan:
## Summary
This PR improves `/v1/messages` observability by preserving the exact model value sent by the client before server-side model mapping, provider alias handling, warmup overrides, or endpoint model normalization. The original requested model is then included inline with the existing upstream Copilot model log.

This makes it easier to diagnose model-mapping behavior from the live server/desktop logs without enabling verbose payload logging or manually correlating handler debug files.

## Before
Only the final upstream model was visible in the live Copilot request log:

```text
<-- model: claude-opus-4.8
```

When a request was mapped from another model, the live log did not show which model the client originally requested.

## After
The same log line now includes both values:

```text
<-- model: claude-opus-4.8 <-- requested model: opus
```

If no rewrite occurs, both fields reflect the same model, which keeps the log format consistent while making mapping behavior explicit.

## Implementation Notes
- Captures the original `/v1/messages` `model` immediately after parsing the request body.
- Carries the requested model through the selected messages flow.
- Applies the inline log format across the Copilot service paths that `/v1/messages` can use: native Messages, Responses, and Chat Completions.
- Leaves direct `/v1/responses` and `/v1/chat/completions` behavior unchanged unless a caller provides requested-model metadata.

## Tests
- `bun test tests/messages-handler.test.ts tests/messages-api-flows.test.ts`
- `bun run lint`
- `bun run typecheck`

### DIFF
--- a/src/routes/messages/api-flows.ts
+++ b/src/routes/messages/api-flows.ts
@@ -81,6 +81,7 @@ export interface FlowBaseOptions {
   requestId: string
   sessionId?: string
   compactType?: CompactType
+  requestedModel: string
 }
 
 interface ResponsesFlowOptions extends FlowBaseOptions {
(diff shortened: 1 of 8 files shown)


caozhiyuan:
@Vivswan npx @jeffreycao/copilot-api@latest start -v and logs in user /.local/share/copilot-api/logs

Vivswan:
Those logs are quite complex. It would be helpful to have a quick model ID / request view for faster diagnostics.

Vivswan:
Also the logs are also difficult to parse because they contain encoded/escaped JSON mixed with other print statements, so `jq` doesn’t work directly either.

Vivswan:
Having the jsonl logs which are easy to parse will work great too

PR #273, opened 2026-06-05
feat: web search for Claude models via GPT /responses backend

mahoshojoHCG:
## Problem

GitHub Copilot offers **no native web search path for Claude models**:

- Copilot's Anthropic-native `/v1/messages` endpoint rejects Anthropic's server-side web search tool: `400 — "The use of the web search tool is not supported." (unsupported_value)` — for every `anthropic-beta` variant and both `web_search_20250305` / `web_search_20260209` tool types.
- Claude models cannot use the `/responses` endpoint at all (`supported_endpoints` is only `["/v1/messages","/chat/completions"]`; a direct call returns `unsupported_api_for_model`).
- Web search **only** works through `/responses`, and **only for GPT models** (confirmed working).

So a `/v1/messages` request that carries a `web_search` server tool currently 400s upstream.

## Solution

Fulfill the web search inside the proxy. When a Claude (Messages API) request carries an Anthropic `web_search` server tool:

1. The server tool is swapped for an equivalent **function tool** Copilot's Claude will call.
2. An agentic loop runs against Copilot's `/v1/messages`; each `web_search` call the model makes is answered by Copilot's **own GPT `/responses` web_search** tool (no external API key, uses existing Copilot quota).
3. The searches + answer are reconstructed into **native Anthropic `server_tool_use` + `web_search_tool_result` blocks** plus the final cited text, so clients render it exactly like real Anthropic web search.

Both **streaming and non-streaming** are supported (streaming replays the reconstructed response as a synthetic Anthropic SSE sequence).

## Config

- `useMessagesApiWebSearch` (default `true`) — enable the feature. When off, the server tool is stripped so requests no longer 400.
- `webSearchBackendModel` (default `"gpt-5-mini"`) — GPT model used as the search backend.

## Notes / limitations

- Mixed assistant turns (a `web_search` call alongside a real client tool in the same response) terminate the loop and hand control back to the client.
- Streaming buffers the loop then replays as SSE — there is inherent latency while searches run, so token-by-token streaming of the final answer is not preserved.

## Testing

- New unit tests in `tests/web-search-fulfill.test.ts` (detection/stripping, fulfillment loop, graceful backend-error block, synthetic stream event sequence).
- `bun run typecheck`, `bun run lint`, and full `bun test` (281 pass) all green.
- Live-verified end to end against Copilot: `claude-sonnet-4.5` invoked `web_search` 3×, each fulfilled via `gpt-5-mini` `/responses`, producing native `server_tool_use` / `web_search_tool_result` blocks, a cited answer, `stop_reason: end_turn`, and `usage.server_tool_use.web_search_requests: 3`.

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -23,6 +23,12 @@ export interface AppConfig {
   useResponsesApiWebSocket?: boolean
   anthropicApiKey?: string
   useResponsesApiWebSearch?: boolean
+  // Enable proxy-fulfilled web search for Claude models on the native
+  // /v1/messages path (Copilot blocks Anthropic's server tool, so the proxy
+  // performs the search itself via the GPT /responses web_search backend).
+  useMessagesApiWebSearch?: boolean
+  // GPT model used as the web search backend for Claude requests.
+  webSearchBackendModel?: string
   claudeTokenMultiplier?: number
 }
 
@@ -121,6 +127,8 @@ const defaultConfig: AppConfig = {
   useMessagesApi: true,
   useResponsesApiWebSocket: true,
   useResponsesApiWebSearch: true,
+  useMessagesApiWebSearch: true,
+  webSearchBackendModel: "gpt-5-mini",
 }
 
 let cachedConfig: AppConfig | null = null
@@ -625,6 +633,16 @@ export function getAnthropicApiKey(): string | undefined {
 export function isResponsesApiWebSearchEnabled(): boolean {
   const config = getConfig()
   return config.useResponsesApiWebSearch ?? true
+}
+
+export function isMessagesApiWebSearchEnabled(): boolean {
+  const config = getConfig()
+  return config.useMessagesApiWebSearch ?? true
+}
+
+export function getWebSearchBackendModel(): string {
+  const config = getConfig()
+  return config.webSearchBackendModel ?? "gpt-5-mini"
 }
 
 export function getClaudeTokenMultiplier(): number {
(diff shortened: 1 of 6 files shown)


caozhiyuan:
Install mcp fetch or other search tools. @mahoshojoHCG

caozhiyuan:
@mahoshojoHCG In Claude Code, when using the WebSearch Tool to invoke a sub-model to initiate a web search, can tools that only use the websearch type be switched to the responses API or message API? Some providers' message APIs also support websearch. There are no plans to support mixing websearch-type tools with other tools, Can you adjust the code? For the switch, configuring a messageApiWebSearchModel should be sufficient.

mahoshojoHCG:
@caozhiyuan Done — pushed two follow-up commits.

  **1. Both switch targets supported, via the one `messageApiWebSearchModel`:**
  - a `provider/model` alias → the `web_search` tool is passed straight through to
    that provider's message API (for an `anthropic`-type provider this is true
    native web search, zero translation);
  - a Copilot GPT model → handled via Copilot's `/responses` web_search.

  The routing decision is a small pure helper (`resolveWebSearchRoute` →
  `provider` / `responses` / `strip`) with unit tests, so it's easy to follow.

  **2. Result fidelity — kept the structured blocks.** Both paths return native
  Anthropic `server_tool_use` + `web_search_tool_result` + answer, so the shape is
  consistent regardless of which target you configure (and explicit source URLs
  survive rather than being buried in markdown).

  **Mixing** stays unsupported: a web_search request with any other tool just has
  the tool stripped and proceeds normally — only web-search-only requests switch.

  typecheck / lint / full suite (287) all pass.

PR #277, opened 2026-06-08
fix: add backoff to Copilot token refresh retries

lostwindsenril:
When running copilot-api in batches, multiple instances can fail to refresh Copilot tokens at the same time and then retry every 15 seconds in lockstep. During transient GitHub/Copilot/network failures, this synchronized retry pattern can leave the batch stuck in collective refresh failures.

Add exponential backoff with jitter for Copilot token refresh retries, cap the retry delay at 10 minutes, and reset the delay after a successful refresh.

批量使用 copilot-api 时发现，多个实例可能同时刷新 Copilot token 失败，并按固定 15 秒节奏集体重试。在 GitHub/Copilot 临时异常、网络抖动或限流时，这种同步重试容易让批量实例陷入集体刷新卡死。

将 Copilot token 刷新失败重试改为带随机抖动的指数退避，最大延迟 10 分钟，并在刷新成功后重置退避延迟。

### DIFF
--- a/src/lib/token.ts
+++ b/src/lib/token.ts
@@ -183,6 +183,8 @@ export const setupCodexToken = async (): Promise<void> => {
 const REFRESH_POLL_INTERVAL_MS = 15_000
 const EARLY_REFRESH_BUFFER_MS = 60_000
 const RETRY_REFRESH_DELAY_MS = 15_000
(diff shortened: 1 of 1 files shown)


caozhiyuan:
@lostwindsenril Why does it freeze?

PR #284, opened 2026-06-10
Stabilize Claude Code billing header cch for better cache hit rate

cuipengfei:
## Summary
- stabilize the volatile `cch` segment in the Claude Code billing header before translating system prompts
- keep unrelated system prompt text unchanged
- add regression tests so translated instructions stay stable and improve Claude Code cache hit rate


caozhiyuan:
@cuipengfei Can it be removed at the message API handler entry?

caozhiyuan:
Or add this env "CLAUDE_CODE_ATTRIBUTION_HEADER": "0"

PR #287, opened 2026-06-11
fix(messages): respect Anthropic cache-control budget

lostwindsenril:
GitHub Copilot's Messages endpoint accepts Anthropic-style cache_control markers, but Anthropic enforces a hard request-wide limit of four cache breakpoints across tools, system, and message content blocks. The existing Messages preprocessing always restored or defaulted a cache_control marker onto the final message block after the tool-result merge pipeline. That preserved cache hints when the merge rewrote the tail block, but it also added a new ephemeral marker when the client did not mark the final message.

This caused valid upstream client requests that already contained four cache_control blocks to be forwarded as five-block requests. A MITM capture of copilot-api's outbound POST /v1/messages showed an incoming four-marker request being sent to api.business.githubcopilot.com with an extra messages[last].content[0].cache_control marker; GitHub Copilot/Anthropic then correctly rejected it with: 'A maximum of 4 blocks with cache_control may be provided. Found 5.' The same payload with three incoming markers was forwarded with four markers and succeeded.

Fix this by treating the automatic final-message marker as budgeted. If the final message did not originally carry cache_control and the request already contains four cache_control blocks, skip adding the default marker. Existing behavior is preserved when there is remaining budget, and explicit cache_control on the original tail is still restored after preprocessing.

Add regression coverage for both sides of the boundary: a four-breakpoint request must not receive a default final-message marker, while a request with remaining budget still gets the automatic final-message cache_control.

### DIFF
--- a/src/routes/messages/preprocess.ts
+++ b/src/routes/messages/preprocess.ts
@@ -247,6 +247,13 @@ export const applyLastMessageCacheControl = (
   anthropicPayload: AnthropicMessagesPayload,
   lastMessageCacheControl: AnthropicCacheControl | undefined,
 ): void => {
+  if (
+    !lastMessageCacheControl
+    && countCacheControlBlocks(anthropicPayload) >= 4
+  ) {
+    return
+  }
+
   const cacheControl = lastMessageCacheControl ?? {
     type: "ephemeral",
   }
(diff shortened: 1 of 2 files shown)


lostwindsenril:
This issue may occur when using Librechat for multi-turn conversations.

image

caozhiyuan:
@lostwindsenril Which client are you using?

caozhiyuan:
@lostwindsenril fixed in 1.11.5

PR #289, opened 2026-06-12
Expose latest Copilot model context limits

lostwindsenril:
## Summary

This PR updates the Copilot Models API version to `2026-06-01` and stops synthesizing separate `[1m]` model IDs in `/v1/models`.

Since GitHub Copilot introduced the new pricing model on `2026-06-01`, most flagship models available to my Copilot Business account, including Claude Opus 4.6-4.8 and GPT-5.4-5.5, expose a 1M context option through the Models API. Fetching this metadata requires the newer `X-GitHub-Api-Version`; the previous API version only returns the smaller/default context window.

image

With the new API response, long context is no longer represented as a separate upstream model. Instead, it is expressed as tiered pricing and context limits under the same model ID, for example:

- default context pricing via `billing.token_prices.default`
- long-context pricing via `billing.token_prices.long_context`
- maximum prompt/context limits via `capabilities.limits`

## Design

This PR intentionally exposes only the upstream-supported maximum context when available, instead of creating local model aliases such as `gpt-5.4[1m]`.

There are two reasons for this:

1. It is a minimal proof of concept that exposes the 1M capability with the smallest possible change.
2. I believe context usage should be controlled by the application/client rather than by the gateway. The gateway should reflect upstream model capabilities and let the caller decide how much context to send.

This also avoids inventing local model IDs that do not exist upstream. The same model ID, such as `gpt-5.4`, is used for both default and long-context usage; the actual pricing tier is determined by the upstream service based on the request's real token usage.

## Changes

- Bump the Copilot API version from `2026-01-09` to `2026-06-01`.
- Remove the `[1m]` suffix generation from `/v1/models`.
- Keep model IDs executable as-is while exposing the latest upstream context metadata.

## Validation

Tested with a Copilot Business account:

- `/v1/models` returns `gpt-5.4` and `gpt-5.5` without synthetic `[1m]` suffixes.
- `gpt-5.4` exposes:
  - `max_prompt_tokens: 922000`
  - `max_context_window_tokens: 1050000`
  - `billing.token_prices.default.context_max: 272000`
  - `billing.token_prices.long_context.context_max: 922000`
- `/v1/responses` with `model: "gpt-5.4"` succeeds.


caozhiyuan:
will supported in next version @lostwindsenril

Issue #286, opened 2026-06-11
seems a bug for the latest version


ghostlitao:
HTTP error: { error:                                                                                                                    3:55:08 PM
   { message:
      'output_config.effort "high" was provided, but model claude-haiku-4.5 does not support reasoning effort',
     code: 'invalid_reasoning_effort' } }

when claude code want to read data using haiku. 

2.1.144 (Claude Code version)

caozhiyuan:
@ghostlitao i'm ok .  what's model config in claude code?

caozhiyuan:
@ghostlitao use claude-haiku-4-5 model id

alexqzd:
Hi @caozhiyuan, I'm getting the same issue in v1.13.3:
```
copilot-api-1  | <-- model: claude-haiku-4.5
copilot-api-1  | <-- model: claude-sonnet-4.6
copilot-api-1  |
copilot-api-1  |  ERROR  Failed to create messages Response {}
copilot-api-1  |
copilot-api-1  |
copilot-api-1  |  ERROR  Error occurred: Failed to create messages
copilot-api-1  |
copilot-api-1  |     at <anonymous> (dist/server-BXCLiIHf.js:5621:13)
copilot-api-1  |     at processTicksAndRejections (native:7:39)
copilot-api-1  |
copilot-api-1  |
copilot-api-1  |  ERROR  HTTP error: { error:
copilot-api-1  |    { message:
copilot-api-1  |       'output_config.effort "high" was provided, but model claude-haiku-4.5 does not support reasoning effort',
copilot-api-1  |      code: 'invalid_reasoning_effort' } }
copilot-api-1  |
copilot-api-1  | --> POST /v1/messages?beta=true 400 560ms
```

This is what i have in my config.json:
```json
  "smallModel": "claude-haiku-4-5",
  "useResponsesApiContextManagement": true,
  "modelResponsesApiCompactThresholds": {
    "gpt-5.4": 217600,
    "gpt-5.5": 217600
  },
  "modelReasoningEfforts": {
    "gpt-5-mini": "low",
    "gpt-5.3-codex": "xhigh",
    "gpt-5.4-mini": "xhigh",
    "gpt-5.4": "xhigh",
    "gpt-5.5": "xhigh"
  },
  "useMessagesApi": true,
  "useResponsesApiWebSocket": true,
  "useResponsesApiWebSearch": true,
```

I also tried setting it to low in modelReasoningEfforts but I get the same error

[...]

PR #296, opened 2026-06-26
fix(messages): don't forward reasoning effort to models that don't support it

timdittler:
## Problem

In `prepareMessagesApiPayload` (`src/routes/messages/preprocess.ts`), reasoning effort is only sanitized **inside** the `adaptive_thinking` branch:

```ts
if (selectedModel?.capabilities.supports.adaptive_thinking && !disableThink) {
  ...
  payload.output_config = { effort }
}
```

Models that don't support `adaptive_thinking` (e.g. `claude-haiku-4.5`, whose `capabilities.supports` declares neither `adaptive_thinking` nor `reasoning_effort`) never enter the branch. So an `output_config.effort` that arrives on the incoming request is forwarded upstream unchanged, and rejected:

```
output_config.effort "high" was provided, but model claude-haiku-4.5 does not support reasoning effort
code: invalid_reasoning_effort
```

This breaks any haiku-backed request carrying a reasoning effort, e.g. lightweight tool-less probes from clients that inherit a session-level effort (Claude Code's `WebFetch` is how I hit it).

## Fix

After the thinking setup, drop `output_config` for **non-adaptive_thinking** models that declare no supported efforts:

```ts
if (
  !selectedModel?.capabilities.supports.adaptive_thinking
  && payload.output_config?.effort
) {
  const supported = selectedModel?.capabilities.supports.reasoning_effort
  if (!supported || supported.length === 0) {
    delete payload.output_config
  }
}
```

Gating on `!adaptive_thinking` leaves adaptive models untouched (they accept an effort, set in the branch above), so the existing fixture expecting `{ effort: "xhigh" }` still holds. The in-branch clamp is unchanged.

## Verification

Built the branch, ran it as the proxy against the Copilot upstream:

| request | result |
|---------|--------|
| `claude-haiku-4-5` + `effort: high` | **200** (effort stripped) |
| `claude-sonnet-4-6` + `effort: high` | **200** (effort kept, thinking active) |
| `claude-haiku-4-5`, no effort | **200** |

Unpatched, the first row returns `400 invalid_reasoning_effort`. `tsc --noEmit` and `eslint` both pass.

Ref: #286

### DIFF
--- a/src/routes/messages/preprocess.ts
+++ b/src/routes/messages/preprocess.ts
@@ -900,4 +900,19 @@ export const prepareMessagesApiPayload = (
       effort: effort,
     }
   }
+
+  // Drop reasoning effort for models that can't use it. A model that supports
+  // adaptive_thinking accepts an effort (handled in the branch above), so this
+  // only targets non-adaptive_thinking models (e.g. claude-haiku-4.5) that also
+  // declare no supported efforts. For those, an effort arriving on the request
+  // would otherwise be forwarded and rejected with `invalid_reasoning_effort`.
+  if (
+    !selectedModel?.capabilities.supports.adaptive_thinking
+    && payload.output_config?.effort
+  ) {
+    const supported = selectedModel?.capabilities.supports.reasoning_effort
+    if (!supported || supported.length === 0) {
+      delete payload.output_config
+    }
+  }
 }


caozhiyuan:
@timdittler There are some other parameters for the output config as well.

PR #299, opened 2026-06-29
feat: translate /chat/completions to Responses API for responses-only models

qxxaa:
### Problem

Newer GPT models from GitHub Copilot (`gpt-5.4-mini`, `gpt-5.4`, `gpt-5.5`) only support the Responses API (`/v1/responses`) and reject `/v1/chat/completions` requests with `unsupported_api_for_model`. The proxy already translates incoming Anthropic `/v1/messages` requests to `/v1/responses` (via `routes/messages/responses-translation.ts`), but the `/v1/chat/completions` endpoint passes requests through to Copilot's `/chat/completions` unchanged - so any OpenAI-format client targeting a responses-only model fails.

This means any tool, agent framework, or application that uses the OpenAI SDK (which always calls `/v1/chat/completions`) cannot use responses-only models through the proxy.

Additionally, the `ChatCompletionsPayload` type only defines `response_format` as `{ type: "json_object" }`, missing the `json_schema` variant needed for grammar-enforced structured output.

### Solution

Add a completions-to-responses translation layer that mirrors the existing Anthropic-to-Responses pattern:

**Routing** (`handler.ts`): After model resolution, check `supported_endpoints` from the live model list. If the model supports `/responses` but NOT `/chat/completions`, route to the new `handleWithResponsesApi` flow. Models with both endpoints continue to use the native `/chat/completions` path.

**Payload translation** (`completions-responses-translation.ts`):
- `messages` -> `instructions` (system/developer) + `input` (user/assistant/tool)
- `response_format.json_schema` -> `text.format` with `strict` preserved
- `ensureAdditionalPropertiesFalse` recursively patches all object nodes (required by Responses API in strict mode)
- `tools` / `tool_choice` / `parallel_tool_calls` -> direct mapping
- `reasoning_effort` -> `reasoning.effort` with `modelReasoningEfforts` config fallback
- `temperature` intentionally omitted (reasoning models reject it)
- `copilot_usage` preserved on non-streaming responses

**Stream translation** (`completions-responses-stream-translation.ts`):
- Responses API events -> `chat.completion.chunk` delta format
- Handles `response.completed`, `response.incomplete`, and `response.failed`
- Guards against undefined `output_index` and `delta` in function call argument events
- Emits proper `[DONE]` termination

**Type extension** (`create-chat-completions.ts`):
- `response_format` type expanded to include `json_schema` with `name`, `schema`, and `strict`

### Files changed

| File | Change |
|---|---|
| `src/routes/chat-completions/completions-responses-translation.ts` | **New.** Payload and result translation. |
| `src/routes/chat-completions/completions-responses-stream-translation.ts` | **New.** Stream event translation. |
| `src/routes/chat-completions/handler.ts` | **Modified.** Responses API routing and flow. |
| `src/services/copilot/create-chat-completions.ts` | **Modified.** Extended `response_format` type. |
| `tests/completions-responses-translation.test.ts` | **New.** 51 test cases for payload and result translation. |
| `tests/completions-responses-stream-translation.test.ts` | **New.** 20 test cases for stream event translation. |

### Verification

- `bun run typecheck` passes (zero errors)
- `bun run lint` passes (zero errors)
- `bun test tests/completions-responses-translation.test.ts tests/completions-responses-stream-translation.test.ts` - 71 tests pass
- `bun test` - full suite passes (429 tests)

### Regression risk

Low. The new code path only activates for models whose `supported_endpoints` does NOT include `/chat/completions`. All existing models that support `/chat/completions` continue on the native path with zero changes. If `state.models` has not loaded yet (cold start edge case), `selectedModel` is undefined, `getResponsesTransportForModel` returns null, and the request falls through to the native path - same behavior as before this change.

### DIFF
--- a/src/routes/chat-completions/completions-responses-stream-translation.ts
+++ b/src/routes/chat-completions/completions-responses-stream-translation.ts
@@ -1,0 +1,259 @@
+/**
+ * Translate Responses API streaming events back to Chat Completions
+ * chunked streaming format (SSE with `chat.completion.chunk` objects).
+ *
+ * This is the completions-side counterpart of the Anthropic-to-Responses
+ * stream translator in routes/messages/responses-stream-translation.ts.
+ */
+
+import type {
+  ChatCompletionChunk,
+  CopilotUsage,
+} from "~/services/copilot/create-chat-completions"
+import type {
+  ResponseOutputItem,
+  ResponsesResult,
+  ResponseStreamEvent,
+  ResponseUsage,
+} from "~/services/copilot/create-responses"
+
+// ---------------------------------------------------------------------------
+// Types
+// ---------------------------------------------------------------------------
+
+interface CompletionsStreamState {
+  responseId: string
+  createdAt: number
+  model: string
+  started: boolean
+}
+
+interface SSEMessage {
+  data?: string
+  event?: string
+}
+
+// ---------------------------------------------------------------------------
+// Public API
+// ---------------------------------------------------------------------------
+
+export const createCompletionsStreamState = (): CompletionsStreamState => ({
+  responseId: "",
+  createdAt: Math.floor(Date.now() / 1000),
+  model: "",
+  started: false,
+})
+
+/**
+ * Translate a single Responses API stream event into zero or more
+ * Chat Completions chunk SSE messages.
+ */
+export const translateResponsesStreamEventToCompletions = (
+  event: ResponseStreamEvent,
+  state: CompletionsStreamState,
+): Array<SSEMessage> => {
+  const messages: Array<SSEMessage> = []
+
+  // Capture metadata from response envelope events
+  if ("response" in event && event.response) {
+    state.responseId = event.response.id
+    state.createdAt = event.response.created_at
+    state.model = event.response.model
+  }
+
+  switch (event.type) {
+    case "response.output_item.added": {
+      handleOutputItemAdded(event.item, event.output_index, state, messages)
+      break
+    }
+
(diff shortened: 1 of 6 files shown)


caozhiyuan:
@qxxaa The Chat Completions API has been deprecated by OpenAI; the Chat interface does not support the pass-through of "reasoning signatures," which negatively impacts the model's intelligence.

caozhiyuan:
you can use gpt model by message api

qxxaa:
Hi @caozhiyuan  Thanks for the feedback! 

Just to clarify - `reasoning_effort` is passed through in the translation (mapped to `reasoning.effort` in the Responses payload), so the model's reasoning depth is preserved. The reasoning signatures (the cryptographic thinking trace) are indeed lost since completions has no field for them, but those are a verification feature rather than something that affects model intelligence.

The main use case here is for clients that only speak `/chat/completions` and currently get a hard 400 when targeting responses-only models like `gpt-5.4-mini`. The translation gives them a working path where they'd otherwise have none.

That said, I understand if this doesn't align with the direction you want for the project. Happy to keep it as a fork if you'd prefer not to add a translation layer for a deprecated API surface.

caozhiyuan:
@qxxaa The chat interface has many, many limitations, and it is completely not recommended to use unless it's an open-source model's chat interface. I suggest you use the gpt response api through the message api, but it's more recommended to directly use the responses api, as many advanced features cannot be converted or translated.

lostwindsenril:
@qxxaa Try LiteLLM

PR #314, opened 2026-07-13
feat(providers): add provider-scoped OpenAI Images proxy

XiaoDcs:
## Problem

Providers may expose native OpenAI-compatible Images endpoints, but copilot-api currently has no provider-scoped path for those requests. Image clients therefore cannot reuse the gateway provider configuration and authentication boundary.

## Solution

- add `POST /:provider/v1/images/generations`
- add `POST /:provider/v1/images/edits`
- add an explicit provider-level `imageEndpoints` allowlist
- stream JSON and multipart request bodies to the native upstream Images endpoints without protocol translation
- preserve multipart boundaries, upstream status codes, response bodies, and compatible response headers
- replace client credentials with the configured provider credential
- preserve `imageEndpoints` when provider credentials are updated through CLI or desktop flows
- document configuration and OpenAI client setup in English and Chinese

## Configuration semantics

`imageEndpoints` declares which native Images paths a provider exposes:

- `["generations"]`: generation only
- `["edits"]`: editing only
- both values: enable both endpoints
- omitted or `[]`: disable Images

This is a provider-level endpoint capability, not a model allowlist. Mixed-model providers are supported: the request `model` is forwarded unchanged and the upstream provider remains responsible for validating model/endpoint compatibility.

The proxy targets `${baseUrl}/v1/images/<endpoint>`; providers with a separate Images base URL should use a separate provider entry. The built-in Codex OAuth provider is intentionally unsupported.

## Security

Gateway client credentials are not forwarded upstream. Requests use the configured provider `apiKey`, while the existing global gateway authentication continues to protect the public route.

## Validation

- provider Images tests: 12/12
- tracked test suite: 431/431 with a 20-second timeout
- TypeScript checks
- main and desktop builds
- relevant main-project ESLint checks


caozhiyuan:
@XiaoDcs for which client?

XiaoDcs:
> @XiaoDcs for which client?

It follows the native OpenAI Images API contract (/v1/images/generations and /v1/images/edits), as used by the official OpenAI SDK. My concrete use case is Codex's Image Gen Skills , but the wire protocol itself is not Codex-specific. The provider prefix is only used for copilot-api routing.

caozhiyuan:
@XiaoDcs which provider support /images/..?

caozhiyuan:
I'm adding the images route for codex gpt-5.6, and will include it in the provider later @XiaoDcs

caozhiyuan:
The provider currently only retains the anthropic interface, I've add it on the top level, and currently only codex is supported.  @XiaoDcs

caozhiyuan:
@XiaoDcs added in latest version

PR #325, opened 2026-07-19
feat(desktop): add launch-at-login option

### LINKED ISSUE #323: [Feature Request] Desktop: launch automatically at OS login
## Problem

The Desktop app does not currently register itself to launch at OS login. After a reboot or sign-in, local clients configured to use Copilot API fail until the app is opened manually.

The existing **Minimize to tray** setting only controls what happens after the app is already running; it does not launch the app at login.

## Proposed behavior

Add an opt-in, default-off setting such as **Launch Copilot API at login**.

- Register or unregister the app using the platform's normal login-startup mechanism.
- When **Minimize to tray** is enabled, start without showing the main window.
- Avoid duplicate app instances.
- Remove or update the startup registration cleanly during uninstall and upgrades.
- Support Windows first, with macOS/Linux where practical.

## Motivation

Copilot API often serves as a local gateway for long-running CLI and editor clients. Those clients should not unexpectedly lose their configured endpoint after a machine restart.

## Environment

- Copilot API Desktop 1.14.10
- Windows

jay-tau:
## Summary

- add an opt-in **Launch at login** setting for packaged Windows, macOS, and Linux desktop apps
- use Electron login items on Windows/macOS and XDG Autostart on Linux, including KDE and GNOME
- start login-launched apps hidden when **Minimize to tray** is enabled
- preserve existing OS login state during migration and respect startup items disabled through OS controls
- keep failed settings saves reversible and resync the settings UI with persisted values on errors
- enforce a single desktop instance and surface the existing window on explicit relaunch
- remove login items when the preference is disabled and clean Windows startup registry entries on uninstall while preserving them during updates
- persist and localize the setting in English and Chinese

## Testing

- `bun test` (457 tests)
- desktop `bun test --coverage tests` (31 tests; `login-item.ts` at 91.67% functions and 99.16% lines; `settings-transaction.ts` at 100%)
- root and desktop TypeScript checks
- root and desktop ESLint checks
- server and Electron production builds
- current Linux x64 AppImage package build and isolated native smoke test proving a `Hidden=true` XDG entry remains byte-for-byte unchanged on startup
- current Windows x64 app bundle build through NSIS staging; final installer emission is host-limited because Wine is unavailable
- current macOS x64 `.app` package build (code signing and DMG creation require macOS)
- earlier native Windows x64 full NSIS install/update/uninstall, startup registry lifecycle, hidden login launch, and single-instance smoke tests
- CachyOS/GNOME Wayland AppImage launch, login-marker launch, and duplicate-instance smoke tests
- native Electron hidden-window show/maximize check
- XDG entry checked by `desktop-file-validate`, `gio`, KDE 6 `kbuildsycoca6`, and GNOME `gtk-launch` process tracing

Native runtime testing passed on Windows x64 build `10.0.26200` and Linux. The review regressions and follow-up edge cases are unit-tested, and the Linux migration behavior is smoke-tested on the current head; refreshed native Windows and native macOS registration tests remain outstanding.

Closes #323


caozhiyuan:
@jay-tau 1. reason=The application rewrites the login item during every initialization. The Windows branch does not pass enabled, which Electron 39.8.10 defaults to true while deleting the StartupApproved disabled value. As a result, if a user disables the startup item through Task Manager or Windows Settings, manually launching the application once enables it again. During migration, normalizing a missing launchAtLogin field to false can also remove an existing same-name macOS or Linux login item; location=desktop/electron/main.ts:332; fix=Register or unregister the login item only when the user explicitly changes the preference. During startup migration, read and preserve the existing OS state; on Windows, preserve the matching launch item's current enabled value when updating its path; occurrences=1
2. reason=settings:save persists the configuration at ipc-handlers.ts:294 before invoking the potentially failing applyLaunchAtLogin call here. On Linux, a valid AppImage path containing a percent sign deterministically throws at login-item.ts:46, and filesystem write or removal errors can also reject. The IPC request then fails after launchAtLogin has already been persisted, while SettingsModal has no error handler, leaving the displayed preference inconsistent with the actual Autostart state; location=desktop/electron/main.ts:346; fix=Make login-item application and settings persistence transactional or rollback to the previous value on failure. Persist launchAtLogin only after registration succeeds and return a visible error to the settings UI; occurrences=1

jay-tau:
@caozhiyuan, I havea addressed the review findings and the two follow-up edge cases:

- preserve the detected OS launch-at-login state even when the desktop settings file is absent
- keep failed settings saves reversible and resync the settings UI with the persisted values

Validation passed: 457 repository tests, 31 desktop tests, lint, typecheck, production builds, coverage, and a fresh Linux AppImage smoke test with the disabled XDG entry and migrated settings unchanged.

PR #333, opened 2026-07-25
fix(messages): preserve Claude Code Auto Mode classifier model

### LINKED ISSUE #332: Claude Code Auto Mode classifier is rewritten to smallModel
## Summary

Claude Code Auto Mode safety-classifier requests are currently treated as tool-less warmup traffic and have their requested model rewritten to `smallModel`.

## Reproduction

1. Point Claude Code at CAO with `ANTHROPIC_BASE_URL`.
2. Enable Claude Code Auto Mode.
3. Configure `smallModel` differently from Claude Code’s classifier model.
4. Trigger an action that needs a safety decision.

Claude Code reports that the classifier model is unavailable and fails closed, blocking the action.

## Root cause

`src/routes/messages/handler.ts` rewrites every non-compact request with an `anthropic-beta` header and no tools to `getSmallModel()`. Auto Mode classifier requests have that shape, but their system prompt contains `<classifier_system_prompt>` and must retain the requested classifier model.

This is separate from the OpenAI Responses API `safety_identifier` field.

## Proposed fix

Exclude requests containing the classifier system-prompt marker from the warmup rewrite. Keep the existing warmup behavior unchanged, and cover both supported `system` representations with a handler regression test.

jay-tau:
## Summary

- preserve the requested model for Claude Code Auto Mode classifier requests
- retain the existing `smallModel` warmup rewrite for ordinary tool-less beta requests
- add regression coverage for string and text-block `system` prompts

Fixes #332

## Testing

- `bun test` — 459 passed
- `tsc --noEmit`
- targeted ESLint on the changed files

### DIFF
--- a/src/routes/messages/handler.ts
+++ b/src/routes/messages/handler.ts
@@ -40,6 +40,21 @@ import { tryHandleWebSearch } from "./web-search/fulfill"
 import consola from "consola"
 
 const logger = createHandlerLogger("messages-handler")
+const CLASSIFIER_SYSTEM_PROMPT_MARKER = "<classifier_system_prompt>"
+
+const isAutoModeClassifierRequest = (
+  system: AnthropicMessagesPayload["system"],
+): boolean => {
+  if (typeof system === "string") {
+    return system.includes(CLASSIFIER_SYSTEM_PROMPT_MARKER)
+  }
+
+  return (
+    system?.some((block) =>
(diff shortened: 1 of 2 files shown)


caozhiyuan:
@jay-tau fixed by feat: skip warmup small-model fallback for copilot provider token-based billing accounts

PR #344, opened 2026-08-06
feat: add Copilot fallback for Codex alpha search

### LINKED ISSUE #311: Web search is unavailable with GPT-5.6 Responses Lite because /alpha/search is not implemented
## Description

Web search is currently unavailable when using GPT-5.6 through copilot-api.

For GPT-5.6 models with `use_responses_lite: true`, Codex exposes the standalone `web.run` tool and sends its commands to `POST /v1/alpha/search`. copilot-api does not currently implement this endpoint, so every web-search attempt returns 404. GPT-5.5 can still search because it uses the hosted `web_search` tool inside `/responses`.

There appear to be two possible approaches:

1. **Configuration workaround:** export a complete custom model catalog, set `use_responses_lite: false` for the GPT-5.6 models, and load it through `model_catalog_json`. Some custom providers may also need `multi_agent_version: "v1"` if they reject the newer namespace-tool schema.

   A related Codex compatibility issue documents this model-catalog workaround and the trade-offs of disabling Responses Lite:

   https://github.com/openai/codex/issues/31870

   This restores the hosted `/responses` web-search path, but disables Responses Lite behavior for those models, may affect standalone tools and multi-agent behavior, and requires maintaining the custom model catalog as Codex models change.

2. **Implement an `/alpha/search` proxy:** translate standalone search requests into the existing Copilot `/responses` `web_search` flow. This would preserve Responses Lite, but full support for follow-up `open`, `click`, and `find` commands would require stable reference IDs and per-session state.

Would you consider supporting the proxy approach, or documenting the configuration workaround?

Falven:
## Summary

- add an opt-in GitHub Copilot Responses fallback for Codex's standalone `/alpha/search` protocol
- recognize every current Codex Search command, including stateful `open`, `click`, and `find`
- preserve authenticated Codex forwarding and provider-scoped route behavior
- return explicit no-retry output for unsupported visual operations and non-live retrieval modes
- add structured references, bounded session state, concise search logging, documentation, and comprehensive tests

## Problem

Codex Responses Lite exposes search as the standalone `web.run` tool and sends
its commands to `POST /alpha/search`. `copilot-api` currently forwards this
endpoint when an authenticated Codex provider is available, but Copilot-provider
users receive a 404.

The current Codex protocol includes search, open, click, find, finance, weather,
sports, time, image search, and PDF screenshots:

https://github.com/openai/codex/blob/57f42a81131ccf5933e7ec5dc659c381eeb5d72b/codex-rs/codex-api/src/search.rs#L8-L305

GitHub Copilot already supports hosted web search through `/responses`, and
`copilot-api` already has Responses search extraction from #273. This PR adapts
that existing path to the standalone Codex protocol without adding another
search service or crawler.

## Solution

When no authenticated Codex provider is available, top-level
`POST /alpha/search` and `POST /v1/alpha/search` can use an opt-in Copilot
fallback.

Routing precedence remains:

1. forward unchanged to an authenticated Codex provider
2. use the enabled Copilot Responses fallback
3. retain the existing 404

Provider-scoped alpha-search routes are unchanged.

The adapter:

- validates the current `SearchRequest` shape with the existing Zod dependency
- retains unknown request fields for forward compatibility
- treats the command payload as authoritative instead of replaying unrelated
  conversation text into the search prompt
- applies model mappings and requires the selected model to support `/responses`
- preserves reasoning, output-token limits, domain filters, approximate location,
  and search-context size
- sends at most one non-streaming Copilot Responses request per alpha-search request
- sets `tool_choice: "required"`, `store: false`, and requests action sources
- records normal Copilot usage and upstream rate-limit behavior
- returns `encrypted_output`, model-visible `output`, and opaque `text_result`
  records containing stable references, URLs, titles, domains, and citation snippets

### Command behavior

| Command | Behavior |
| --- | --- |
| `search_query` | Runs Copilot hosted search and assigns stable `turn{n}search{i}` references |
| `open` | Resolves a reference or literal URL, returns line-numbered grounded text, and stores a bounded view snapshot |
| `click` | Resolves a numbered link from an opened view and handles it as `open` |
| `find` | Searches cached view text locally first, then asks Copilot for matching page context |
| `finance` | Translates ticker, asset type, and market constraints to hosted search |
| `weather` | Translates location, start date, and duration to hosted search |
| `sports` | Translates league, schedule/standings, team, opponent, date, and locale constraints |
| `time` | Computes the requested UTC-offset time locally without invoking Copilot |
| `image_query` | Returns successful unsupported/no-retry output because Copilot web search has no image-search result protocol |
| `screenshot` | Returns successful unsupported/no-retry output because Copilot cannot return PDF page screenshots |
| unknown future commands | Return successful unsupported/no-retry output instead of causing retry loops |

Mixed requests execute supported commands and append one warning per unsupported
command type.

Explicit `external_web_access: false`, `cached`, or `indexed` requests are never
silently upgraded to live retrieval.

### Reference state

References use bounded process-local state keyed by `SearchRequest.id`:

- sliding one-hour session TTL
- maximum 128 sessions
- maximum 256 URL references per session
- maximum 16 opened-page snapshots per session
- maximum 20,000 characters per snapshot
- lazy expiry and oldest-entry eviction, with no background timer
- synchronous turn reservation to avoid duplicate IDs during concurrent requests
- URL deduplication with stable references

Restarting the proxy intentionally invalidates this ephemeral state. Missing or
expired references return recoverable output instructing Codex to search or open
the URL again.

Copilot performs all remote retrieval. The proxy does not fetch arbitrary URLs,
which avoids adding a crawler, HTML parser, dependency, or new SSRF boundary.

## Configuration

The fallback is disabled by default and requires Responses web search to remain
enabled:

```json
{
  "useResponsesApiWebSearch": true,
  "alphaSearchResponsesFallback": true
}
```

Codex must be configured to use standalone live search:

```toml
web_search = "live"

[model_providers.copilot_api]
supports_standalone_web_search = true
```

README documentation is included in English and Chinese.

## Observability

Completed remote searches emit one concise line containing the operation types,
queries Copilot actually executed, and the number of returned sources:

```text
--> web search: operations=search_query queries=["actual query"] sources=1
```

Answer text, snippets, and URLs are not written to the normal terminal log.

## Limitations

- the fallback supports live retrieval only
- `image_query` and `screenshot` are recognized but cannot be fulfilled by
  GitHub Copilot web search
- references are intentionally process-local and expire after one hour
- opened-page content is Copilot-grounded text, not locally fetched raw HTML
- the feature remains default-off while compatibility is evaluated

## Validation

Automated verification:

- `bun test tests/alpha-search.test.ts` — 28 passed
- `bun test tests/web-search-fulfill.test.ts` — 15 passed
- `bun run lint:all` — passed
- `bun run typecheck` — passed
- `bun test` — 491 passed, 0 failed
- `bun run build` — passed
- `git diff --check` — passed
- core changed search modules have 97.19–100% line coverage

Live verification with `gpt-5.6-sol`:

- direct `/alpha/search` returned cited structured references
- Codex invoked standalone Search successfully
- `search_query` produced stable source references
- `open` produced a numbered view with resolvable links
- cached `find` returned immediately without another Copilot request
- query text and source counts appeared in the new concise log output

No dependencies or version changes are included.

Fixes #311

### DIFF
--- a/README.md
+++ b/README.md
@@ -561,6 +561,7 @@ Use `copilot-api auth login --provider custom` to add or update another third-pa
     "useMessagesApi": true,
     "useResponsesApiWebSocket": true,
     "useResponsesApiWebSearch": true,
+    "alphaSearchResponsesFallback": false,
     "messageApiWebSearchModel": "gpt-5-mini"
   }
   ```
@@ -648,6 +649,15 @@ Use `copilot-api auth login --provider custom` to add or update another third-pa
 - **useMessagesApi:** When `true`, Claude-family models that support Copilot's native `/v1/messages` endpoint will use the Messages API; otherwise they fall back to `/chat/completions`. Set to `false` to disable Messages API routing and always use `/chat/completions`. Defaults to `true`.
 - **useResponsesApiWebSocket:** When `true`, Responses API requests use Copilot's websocket transport for models that advertise `ws:/responses`; models that only advertise `/responses` continue to use HTTP. Set to `false` to disable websocket routing and use HTTP `/responses` whenever the selected model supports it. Defaults to `true`. If the Responses API WebSocket gets closed, it is usually caused by your own network. If you are using a VPN, try switching to a different node.
 - **useResponsesApiWebSearch:** When `true`, the server keeps Responses API tools with `type: "web_search"` and forwards them upstream. Set to `false` to strip those tools from `/responses` payloads. Defaults to `true`.
+- **alphaSearchResponsesFallback:** When `true` and `useResponsesApiWebSearch` remains enabled, top-level `POST /alpha/search` and `POST /v1/alpha/search` requests use GitHub Copilot Responses web search when no authenticated Codex provider is available. An authenticated Codex provider still takes precedence, and provider-scoped alpha-search routes are unchanged. The fallback recognizes every current Codex search command; unsupported `image_query` and `screenshot` operations return successful no-retry tool output. Defaults to `false`. Configure Codex to call the endpoint with:
+
+  ```toml
+  web_search = "live"
+
+  [model_providers.copilot_api]
+  supports_standalone_web_search = true
+  ```
+
 - **messageApiWebSearchModel:** Global fallback model used when a top-level Copilot `/v1/messages` request contains only the server-side `web_search` tool. Defaults to `gpt-5-mini`. If the value is a `provider/model` alias, the request is routed into that provider's Messages API path with the provider prefix stripped. For Copilot GPT models, web search runs through `/responses`. Mixed `web_search` plus custom tools are not supported and the server-side `web_search` tool is stripped.
 - **claudeAutoModel:** Model used for Claude Code background security-monitor requests on `/v1/messages` and provider message routes. A request is treated as a security-monitor request when it carries no tools, sets `stop_sequences` to `["</block>"]`, and contains a system text block starting with `You are a security monitor for autonomous AI coding agents.`; its model is then replaced with this value. For top-level requests, a `provider/model` alias is forwarded into that provider's Messages API; provider routes keep their current provider and use this configured value directly. Defaults to empty (disabled).
 - **claudeTokenMultiplier:** Multiplier applied to the fallback GPT-tokenizer estimate for Claude `/v1/messages/count_tokens` requests. Defaults to `1.15`. Increase it if your client is still compacting too late. This setting is only used when the proxy is estimating Claude tokens locally; if `anthropicApiKey` is configured and Anthropic token counting succeeds, the exact Anthropic count is returned instead.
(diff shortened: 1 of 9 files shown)


caozhiyuan:
Is there official documentation for the serach interface? What is the logic behind it? @Falven

Falven:
> Is there official documentation for the serach interface? What is the logic behind it? @Falven

There isn’t a complete public API reference for /alpha/search; it’s an alpha Codex interface. The authoritative contract is the open-source [SearchRequest schema](https://github.com/openai/codex/blob/57f42a81131ccf5933e7ec5dc659c381eeb5d72b/codex-rs/codex-api/src/search.rs#L8-L305). OpenAI’s [app-server documentation](https://github.com/openai/codex/blob/main/codex-rs/app-server/README.md#L1580) only briefly mentions its structured results.
This implementation is an adapter, not a search engine. It forwards requests unchanged when Codex authentication is available; otherwise it translates supported commands into Copilot Responses hosted web search, maintains bounded references for open/click/find, computes time locally, and converts the answer and sources back into Codex’s response format. Unsupported visual or non-live operations return explicit no-retry results. Copilot performs all remote retrieval and the proxy does not crawl URLs itself.

PR #354, opened 2026-08-12
Harden Responses transport lifecycle and buffering

es2737:
## Why

I use `copilot-api` every day and have been experiencing several reliability issues around cancelled or stalled Responses requests, especially with long-running streams and the pooled WebSocket transport. In practice, client disconnects could leave upstream generations running, active streams could hang indefinitely, and slow consumers could grow the in-memory WebSocket queue without a bound.

This PR is my attempt to fix those issues comprehensively. I would appreciate a review even if you would prefer to reimplement some or all of the approach in a different shape.

## What changed

- propagate each downstream request's `AbortSignal` through the native GitHub Copilot, ChatGPT Codex-provider, and generic provider Responses paths
- return a quiet 499 for client cancellation and deterministically release body readers, stream iterators, timers, listeners, and request-specific sockets
- add distinct configurable HTTP headers, stream inactivity, WebSocket open, active-stream inactivity, and reusable-pool idle timeouts; there is no short total-generation deadline
- bound ordered WebSocket buffering by both bytes and message count, failing and invalidating the affected socket rather than dropping deltas
- prevent concurrent or late WebSocket events from crossing request boundaries, while preserving reuse after a clean terminal response
- invoke the existing unsupported-input sanitizer on the actual GitHub Copilot HTTP and WebSocket path, without applying Copilot-specific stripping to the separate ChatGPT Codex backend or generic providers
- correct the built-in GPT-5.6 Sol/Terra/Luna reasoning metadata (`none` through `max`, with raw `ultra` kept separate) and context/output limits
- remove order-dependent alpha-search tests caused by process-wide Bun module mocks, replacing them with scoped dependency overrides that are restored after each test
- preserve cache-relevant payload ordering, reasoning/encrypted content, tool IDs, `prompt_cache_key`, and backend-specific request behavior

The default limits are conservative and backward compatible:

- HTTP headers: 30 seconds
- active HTTP/WebSocket inactivity: 5 minutes
- WebSocket open: 30 seconds
- reusable WebSocket pool idle: 60 seconds
- WebSocket buffer: 8 MiB or 1,024 messages

All values are configurable under `responsesTransport` and documented in both READMEs.

## Validation

- `bun run typecheck`
- `bun run lint:all`
- `bun run build`
- targeted HTTP, WebSocket, handler, provider, sanitizer, model, cache-stability, and alpha-search tests
- repeated full-suite runs; final result: 597 passed, 0 failed across 61 files
- changed executable source-line coverage: 93.29% (709/760)

All upstream behavior was exercised with mocks and local fixtures; no live GitHub Copilot, ChatGPT backend, or paid OpenAI request was made during validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -649,6 +649,14 @@ Use `copilot-api auth login --provider custom` to add or update another third-pa
     },
     "useMessagesApi": true,
     "useResponsesApiWebSocket": true,
+    "responsesTransport": {
+      "headersTimeoutMs": 30000,
+      "streamInactivityTimeoutMs": 300000,
+      "websocketOpenTimeoutMs": 30000,
+      "websocketPoolIdleTimeoutMs": 60000,
+      "websocketMaxBufferedBytes": 8388608,
+      "websocketMaxBufferedMessages": 1024
+    },
     "useResponsesApiWebSearch": true,
     "alphaSearchCodexPriority": true,
     "alphaSearchModel": "gpt-5-mini",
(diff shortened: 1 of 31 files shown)


caozhiyuan:
@es2737  gpt-5.6-sol review result : [HIGH_PRIORITY]

  1. Codex HTTP SSE body is not cancelled when the consumer stops at a terminal event

     createResponsesSafeStream(events(response)) does not cancel or release the underlying reader when the consumer exits after response.completed or
     response.incomplete. A minimal reproduction returned cancelled=0 and bodyLocked=true. If the upstream server keeps the HTTP body open after the terminal
     event, the reader, lifecycle listeners, and HTTP connection may remain allocated.

     Location: src/services/codex/create-responses.ts:284

     Fix: Own the response reader in the Codex HTTP stream wrapper and cancel/release it in a finally block. Add a regression test where the upstream emits a
     terminal event without closing the body, then assert that the body is cancelled and unlocked.

es2737:
@caozhiyuan Fixed, Codex HTTP SSE streams now own the response reader and always cancel and release it in a `finally` block when consumption ends, including after a terminal event while the upstream body remains open.

I also added a regression test that emits `response.completed` without closing the upstream body and verifies it is cancelled and unlocked. All 609 tests, lint, root/desktop typechecks, and build pass locally.

es2737:
I use this project everyday in Codex as my main subscription, but it's a bit unreliable for longer running tasks as it may timeout or fail on encryption so I aim to solve it. I'm happy to contribute more! @caozhiyuan

caozhiyuan:
@es2737 Can it be merged? and ci falied

es2737:
Yes, it is mergeable with no conflicts. The previous CI failure was caused by a timing-sensitive cancellation test, which I’ve fixed in the latest commit (`7c85030`). I verified it locally with 26 repeated targeted runs and the full suite: 609 passed, 0 failed.  

The new CI run is currently marked “action required” because this is a fork PR. Could you please approve and run the workflow?  @caozhiyuan

caozhiyuan:
@es2737 test timeout 5s

PR #367, opened 2026-08-21
fix: validate provider existence before routing provider/model aliases

lenonis:
## Problem

`parseProviderModelAlias` treats any model name containing `"/"` as a custom provider alias (e.g. `"dash/qwen-plus"`). Some model ids are namespaced (e.g. `"org/family/model"`) where the prefix is **not** a configured provider. Such requests were misrouted to a non-existent provider and surfaced as a `400` error (`Provider '<name>' does not support the /v1/<...> endpoint`) instead of falling through to the default model lookup.

## Fix

Add a `resolveProviderConfig` check in the **chat-completions**, **messages**, and **responses** route handlers so the `"/"` prefix is only treated as a custom provider alias when that provider is actually configured. When the prefix does not match a configured provider, the request falls through to the default model lookup as intended.

`parseProviderModelAlias` itself is left as a pure string parser. The existence check lives at the route-handler call sites so it stays consistent with the existing per-route dependency seams:

- `chat-completions` route uses the directly-imported `resolveProviderConfig` (its tests already mock `getProviderConfig`, which `resolveProviderConfig` falls through to for non-codex providers).
- `responses` route uses the existing `responsesHandlerDependencies` seam (the dependency object now exposes `resolveProviderConfig`).
- `messages` route uses the existing `providerMessagesHandlerDependencies.resolveProviderConfig` seam.

## Why not patch `parseProviderModelAlias` directly?

Putting the existence check inside `parseProviderModelAlias` (e.g. via `getRawProviderConfig`) would work for the chat-completions path, but it breaks other call sites that route codex/`<provider>` aliases through dependency-injected `resolveProviderConfig` rather than the raw config store — the two lookups disagree in test environments where the codex provider is configured via `persistCodexCredentials`. Keeping `parseProviderModelAlias` purely syntactic and doing the existence check at the route handlers avoids that mismatch and matches the existing seam design.

## Tests

- New regression test in `tests/provider-model-alias.test.ts` verifying a namespaced id with an unconfigured prefix (`"org/family/model"`) is **not** routed to the provider path and does not surface the provider `400` wording.
- Updated `tests/provider-responses-context-management.test.ts` and `tests/provider-model-alias.test.ts` config mocks to be name-aware so the new existence check reflects real config-store semantics.

```
$ bun test
682 pass, 0 fail, 2316 expect() calls   (68 files)

$ npx eslint <changed files>
clean
```

## Notes

This was motivated by real-world usage where namespaced model ids (custom enterprise models exposed via the GitHub Copilot gateway) were rejected with a `400` when sent through this proxy. The fix is generic and does not reference any specific model or provider name.

### DIFF
--- a/src/routes/alpha-search/route.ts
+++ b/src/routes/alpha-search/route.ts
@@ -222,10 +222,21 @@ export async function handleAlphaSearchRequest(
     messagesBackedModel ? payload.model : requestedModel
 
   if (providerModelAlias) {
-    return await alphaSearchRouteDependencies.handleAlphaSearchResponses(c, {
-      provider: providerModelAlias.provider,
-      request: payload,
-    })
+    // Only treat the "/" prefix as a custom provider alias when that provider
+    // is actually configured. Namespaced model ids such as "org/family/model"
+    // must fall through to the default alpha search path instead of being
+    // passed to handleAlphaSearchResponses with an unconfigured provider and
+    // surfaced as a 404.
+    const providerConfig =
+      await alphaSearchRouteDependencies.resolveProviderConfig(
+        providerModelAlias.provider,
+      )
+    if (providerConfig) {
+      return await alphaSearchRouteDependencies.handleAlphaSearchResponses(c, {
+        provider: providerModelAlias.provider,
+        request: payload,
+      })
+    }
   }
 
   return await alphaSearchRouteDependencies.handleAlphaSearchResponses(c, {
--- a/src/routes/chat-completions/handler.ts
+++ b/src/routes/chat-completions/handler.ts
@@ -7,6 +7,7 @@ import { resolveMappedModel } from "~/lib/config"
 import { createHandlerLogger, debugJson } from "~/lib/logger"
 import { findEndpointModel } from "~/lib/models"
 import { parseProviderModelAlias } from "~/lib/provider-model"
+import { resolveProviderConfig } from "~/lib/provider-resolver"
 import {
   createCopilotTokenUsageRecorder,
   normalizeOpenAIUsage,
@@ -36,11 +37,21 @@ export async function handleCompletion(c: Context) {
 
   const providerModelAlias = parseProviderModelAlias(payload.model)
   if (providerModelAlias) {
-    payload.model = providerModelAlias.model
-    return await handleProviderChatCompletionsForProvider(c, {
-      payload,
-      provider: providerModelAlias.provider,
-    })
+    // Only treat the "/" prefix as a custom provider alias when that provider
+    // is actually configured. GitHub Copilot enterprise models can ship with
+    // namespaced ids such as "org/family/model" that must fall through to the
+    // default model lookup instead of being misrouted to a non-existent
+    // provider and surfaced as a 400.
+    const providerConfig = await resolveProviderConfig(
+      providerModelAlias.provider,
+    )
+    if (providerConfig) {
+      payload.model = providerModelAlias.model
+      return await handleProviderChatCompletionsForProvider(c, {
+        payload,
+        provider: providerModelAlias.provider,
+      })
+    }
   }
 
   debugJson(logger, "Request payload:", payload)
--- a/src/routes/chat-completions/handler.ts
+++ b/src/routes/chat-completions/handler.ts
@@ -42,6 +42,12 @@ export async function handleCompletion(c: Context) {
     // namespaced ids such as "org/family/model" that must fall through to the
     // default model lookup instead of being misrouted to a non-existent
     // provider and surfaced as a 400.
+    //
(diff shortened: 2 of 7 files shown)


caozhiyuan:
@lenonis it's copilot models ?

caozhiyuan:
Thanks for the fix — the approach (keeping `parseProviderModelAlias` purely syntactic and checking provider existence at the route handlers) is sound and consistent with the existing DI seams. A few things before merging:

**Same class of bug still uncovered at other `parseProviderModelAlias` call sites:**

1. `src/routes/messages/count-tokens-handler.ts` — a namespaced model id sent to `/v1/messages/count_tokens` is still routed to `handleProviderCountTokensForProvider` and returns 404 `Provider '<name>' not found or disabled` instead of falling through to the Anthropic/tokenizer-estimation path. This matters in practice: clients like Claude Code call `count_tokens` before the actual request, so the main flow would work while the preflight call keeps failing.
2. `src/routes/alpha-search/route.ts` — a namespaced id still ends up in `handleAlphaSearchResponses` with `provider: <unconfigured prefix>`.

Could you either cover these here or open a follow-up issue so they don't get lost?

**Minor nits (non-blocking):**

- Valid aliases now pay for two `resolveProviderConfig` calls (route handler + provider handler); for `codex` that means `setupCodexToken()` runs twice. Idempotent, just slightly wasteful.
- chat-completions imports `resolveProviderConfig` directly while the other two routes go through DI seams — the justification in the PR body makes sense, but a short comment at the call site would help future readers.
- The new regression test only covers `/v1/messages`; the chat-completions and responses fallthrough paths have no test. The assertions are also all negative (not routed to dash, error wording absent) — an assertion that the request actually reaches the default model lookup would make the test stronger.

**One request:** could you share a concrete, real-world namespaced model id that the GitHub Copilot gateway actually returns (e.g. from the `/models` listing of an enterprise account)? The PR says the fix "does not reference any specific model", which is fine for the code, but having a real example in the PR description or a test comment would make the scenario much easier to verify and guard against regressions.

lenonis:
Thanks for the thorough review — pushed an update (242aa19) addressing the call sites you flagged and the test/doc nits.

## Covered call sites

**`src/routes/messages/count-tokens-handler.ts`** — the same `resolveProviderConfig` guard is now applied. A namespaced id sent to `/v1/messages/count_tokens` now falls through to the Anthropic/tokenizer-estimation path instead of 404ing via `handleProviderCountTokensForProvider`. New test asserts the estimation path is reached (200, `getTokenCount` called) for a namespaced id, plus a happy-path guard that a configured alias still routes to the provider token counter.

**`src/routes/alpha-search/route.ts`** — the final `if (providerModelAlias)` block now checks `alphaSearchRouteDependencies.resolveProviderConfig` before delegating to `handleAlphaSearchResponses` with a `provider:`. An unconfigured prefix falls through to the default (no-provider) alpha search path instead of 404ing inside `resolveRemoteModel`. (Uses the existing DI seam, consistent with `isMessagesBackedModel` right below it.)

## Nits

- **Double `resolveProviderConfig` / `setupCodexToken`**: acknowledged and left as a follow-up. Eliminating the second call means refactoring the provider handler signatures to accept a pre-resolved `ResolvedProviderConfig` (they currently take a `provider: string`), which felt like scope creep here. `setupCodexToken()` is idempotent so behavior is correct, just slightly wasteful. Happy to do it as a separate PR if you want.
- **chat-completions direct import**: added a comment at the call site explaining the asymmetry (no route-level DI seam here; existing tests mock `getProviderConfig`, which `resolveProviderConfig` falls through to for non-codex providers).
- **Test coverage**: the new `count_tokens` test has positive assertions (estimation path reached, 200). The `/v1/messages` test keeps a negative assertion since its default flow needs a real Copilot token to fully succeed; I can add a full positive test there too if you want, but it would mean mocking the Copilot token flow.

## Real-world namespaced id

Good call — adding a concrete example. The GitHub Copilot gateway exposes enterprise-configured models under the listing as `account/model` and, for nested org-scoped models, `account/family/model` (the enterprise admin configures a model `slug` that gets prefixed with the org account handle). For example, an enterprise account `contoso` exposing a custom `glm-5.2` returns a `/models` entry whose `id` is literally `contoso/glm-5.2` (single-segment namespacing), and deeper org-scoped entries can look like `contoso/family/glm-5.2` (multi-segment). Both shapes trip the bug: `parseProviderModelAlias` treats `contoso` as a provider alias prefix, but there is no `contoso` entry in `config.providers`, so the request 400/404s at the provider handler instead of being sent upstream as-is.

I will add a short test comment capturing this real shape in the next push so the regression guard is self-documenting. Let me know if the example format above is accurate for what you have seen; I based it on the `/models` listing shape returned by the gateway for enterprise accounts.

PR #373, opened 2026-08-26
security: add opt-in hardening with compatible defaults

NaveDanan:
Replaces #371 after addressing the maintainer's backward-compatibility feedback.

## Why

The gateway handles API keys, provider tokens, prompts, and model responses. The first revision applied the new security policy by default, which could break existing LAN, browser, Docker, provider, and Usage Viewer setups.

This revision keeps existing behavior by default. Users do not need to change their configuration after upgrading. The stricter policy is now opt-in.

## Backward compatibility

Without increased security enabled, the gateway keeps the previous behavior:

- listens on `0.0.0.0` unless `--host` is set
- keeps wildcard CORS and remote Usage Viewer endpoints
- keeps the `ANTHROPIC_API_KEY` fallback
- keeps existing provider and Copilot endpoint routing
- keeps legacy Zstandard request handling
- keeps the existing Docker `GH_TOKEN` startup path

No existing CLI flags or configuration files need to change.

## Increased security mode

Users can enable the hardened policy with `--strict-security`.

The desktop app adds an **Increased security** switch in Settings > Security. It is off by default. On first use, the app shows a suggestion dialog with **Enable increased security** and **Not now** buttons. The app saves the choice before the first manual or automatic server start.

When enabled, strict mode:

- defaults to the loopback listener
- validates Host and browser Origin headers
- requires authentication and protected transport for remote access
- validates provider and Copilot endpoints
- restricts the Usage Viewer to same-origin requests and session-only credentials
- applies compressed and expanded Zstandard body limits
- protects Docker credentials from process arguments

Security changes that do not alter supported workflows, such as constant-time key comparison, credential redaction, Electron IPC sender checks, and safer logging, remain enabled in both modes.

## Validation

- `bun test`: 786 passed
- `bun run --cwd desktop test`: 46 passed
- root and desktop typechecks passed
- root, desktop-server, and Electron production builds passed
- Bash syntax validation passed for `entrypoint.sh`
- desktop dialog and Settings switch were checked at desktop and compact window sizes


caozhiyuan:
@NaveDanan Please split into smaller PRs, and do not make changes that are optional

PR #375, opened 2026-08-27
feat: support discovered fast model routing

### LINKED ISSUE #374: Support routing fast tiers to separate model IDs
Copilot exposes some fast variants as separate models. For example, Claude Opus 4.8 Fast uses a distinct model ID rather than Anthropic's `service_tier`, which controls capacity priority and is unrelated to fast inference.

Add generic tier-aware routing when a separate fast model exists:

```text
claude-opus-4.8 + service_tier=priority -> claude-opus-4.8-fast
```

Standard requests and models without a fast variant should remain unchanged. Apply this consistently to HTTP and WebSocket requests.

jay-tau:
## Summary
- discover available `-fast` model variants from the Copilot model catalog
- route Codex `service_tier: priority` and Claude Code `speed: fast` requests before capability normalization
- advertise fast tiers only when a compatible variant exists and configure Claude Code gateway support
- surface upstream Responses stream failures as SSE errors

Fixes #374

## Validation
- `bun test`
- `bun run lint:all`
- `bun run typecheck`
- `bun run build`


caozhiyuan:
Thanks for the work here — the exploration of Copilot's `-fast` catalog variants is genuinely useful, and the tests document the intended behavior well. After reviewing, though, I've decided not to merge this in its current form. My reasons:

**1. The core value is already achievable without gateway changes.**
Users who want a fast model can pin the `-fast` model id directly today — via `ANTHROPIC_MODEL` / model config or an existing `modelMappings` entry. The incremental value of this PR is making the native `/fast` toggle (`speed: "fast"`) and Codex `service_tier: "priority"` work through the gateway with zero configuration. That's real, but modest, convenience — and it comes at the cost of new request-time routing logic that every message and responses request now passes through.

**2. It adds a scattered, evolving business rule to the hot path.**
The `${id}-fast` naming convention and the "when is a fast variant compatible" rule are implemented three times with three different standards: the `/v1/models` advertisement checks `isCopilotCodexCandidate` + transport, the Responses route checks transport/messages-fallback, and the Messages route checks nothing at all. If Copilot changes the naming or endpoint layout, three call sites must change in lockstep or the gateway silently diverges. This belongs behind a single resolver (ideally computed once at catalog refresh, not per request), and I'd want that refactoring before taking on the feature.

**3. The SSE `onError` additions are a net regression.**
hono's `streamSSE` already emits its own `event: error` frame after a custom `onError`, so these handlers produce two error events per failure, the second being non-JSON. Meanwhile the failure mode they target is already covered twice: hono logs the error server-side, and clients already retry when no terminal event (`response.completed` / `message_stop`) arrives. The added code buys little and introduces a deterministic protocol bug.

**4. Unverified client-side assumptions.**
The behavior of Codex and Claude Code when receiving a generic SSE `error` frame (vs. a protocol-shaped `response.failed`) is untested, and the catalog-dependent behaviors (which models get `-fast` variants, with which endpoints) can't be validated by unit tests alone.

If there's interest in pursuing this later, I'd be glad to review a version that: drops the SSE changes entirely, centralizes fast-variant discovery/compatibility into one helper computed at catalog load, and applies the same guard standard on all three paths. Closing for now — thank you again for the effort.

PR #379, opened 2026-09-02
fix(responses): isolate reasoning by transport

### LINKED ISSUE #378: bug: Responses and Messages model switches replay incompatible reasoning
## Summary

A multi-turn `/v1/responses` conversation can fail when the client switches between a native Responses model and a Messages-backed model while replaying prior output.

The observed failure is GPT → Claude. GPT returns a native reasoning item with `encrypted_content`; the Messages fallback translates that opaque value into an Anthropic `thinking.signature`. Claude rejects the foreign signature:

```text
messages.N.content.N: Invalid signature in thinking block
```

## Root cause

The public Responses history uses the same reasoning item shape for two different upstream transports:

- native Responses reasoning
- Claude reasoning synthesized by Messages → Responses translation

On replay, the gateway did not distinguish those two sources before translating or forwarding the next request. Native Responses reasoning could therefore be treated as a Claude signature, and synthetic Claude reasoning could be sent to a native Responses target.

This issue only claims the transport boundary demonstrated by the failing path. It does not require provider- or model-specific provenance.

## Reproduction

1. Send a `/v1/responses` request to a native Responses model with `reasoning.encrypted_content` included.
2. Replay the returned output in the next turn.
3. Switch to a Messages-backed Claude model.
4. The native opaque value may be translated into `thinking.signature` and rejected.

The reverse switch has the symmetric risk: a synthetic Claude reasoning item can be forwarded to native Responses.

## Expected behavior

- Native Responses targets receive only native, unmarked reasoning items.
- Messages-backed targets receive only reasoning synthesized by the Messages adapter.
- Visible assistant messages, user messages, tool calls, tool results, and compaction items remain unchanged.
- Filtering affects only the current upstream request; client-retained history remains complete, so switching back can replay the original transport's reasoning again.
- Native reasoning IDs and all opaque `encrypted_content` values remain byte-for-byte unchanged.

## Proposed scope

Mark only reasoning IDs synthesized by Messages → Responses with a reserved suffix. Once the actual transport is selected, filter replayed reasoning by that marker:

- Messages fallback: keep marked reasoning, drop unmarked reasoning
- native Responses: keep unmarked reasoning, drop marked reasoning

Legacy unmarked Claude reasoning cannot be distinguished safely and is dropped on its next Messages-backed replay.

Provider-prefixed routes, direct `/v1/messages`, `/chat/completions`, warmup behavior, provider proxies, logging, and custom-tool item IDs are out of scope.

## Cache tradeoff

Switching models may miss the prompt cache on the first request, especially for very large sessions. That is an expected cost of model switching. The gateway should provide a valid switch and preserve visible history, but it cannot guarantee cross-model cache continuity.

LaChimere:
Closes #378

## Summary

Fix model switching in multi-turn `/v1/responses` conversations without introducing a general reasoning carrier.

The failure occurs when native Responses reasoning is replayed through the Messages fallback and its opaque `encrypted_content` is treated as an Anthropic `thinking.signature`. Claude rejects that foreign signature. The reverse switch can likewise send synthetic Claude reasoning to a native Responses target.

## Fix

This PR follows the transport marker approach suggested in review:

- append the reserved `__a1` suffix only to reasoning IDs synthesized by Messages → Responses
- keep native Responses reasoning IDs and `encrypted_content` unchanged
- after transport selection, filter only `type: "reasoning"` items in the current upstream request
  - Messages fallback keeps marked reasoning
  - native Responses keeps unmarked reasoning
- leave visible messages, tool calls, tool results, and compaction items untouched
- retain complete client history, allowing `GPT → Claude → GPT` and `Claude → GPT → Claude` round trips

Legacy unmarked Claude reasoning is dropped once because it cannot be distinguished safely from native reasoning.

## Scope

The implementation is intentionally limited to five files: three production files and two focused test files.

It does not change:

- opaque `encrypted_content`
- native reasoning IDs
- provider-prefixed Responses routes
- direct `/v1/messages` or `/chat/completions`
- warmup/model routing
- provider proxy headers or body rewriting
- logging
- compaction format
- custom-tool item IDs

The existing synthesized terminal SSE response continues to use `output: []`; SSE marker consistency is verified on the added/done item events.

## Cache behavior

This change makes switching valid; it does not promise a cache hit across models. A first request after switching can miss the prompt cache, particularly in a very large session. That cost is an inherent product tradeoff of model switching rather than something the gateway should hide or emulate.

Within the same transport category, reasoning is preserved and the upstream remains responsible for cache compatibility.

## Tests

Focused tests cover:

- GPT/native → Claude/Messages filtering
- Claude/Messages → GPT/native filtering
- three-request `GPT → Claude → GPT` round trip
- three-request `Claude → GPT → Claude` round trip
- replay of the actual prior `response.output` under one `session-id`
- same-transport reasoning preservation
- visible assistant messages, function calls, and tool results
- non-stream Messages reasoning IDs and unchanged signatures
- SSE added/done events using the same marked reasoning ID
- legacy unmarked reasoning being removed from Messages requests

Validation for commit `9e2baea`:

- focused tests: 75 passed, 0 failed
- full `bun test`: 762 passed, 0 failed
- `bun run typecheck`: passed
- changed-file lint fix and full read-only lint: passed
- `bun run build`: passed
- `git diff --check`: passed

The focused coverage run directly exercises both new filtering branches and both JSON/SSE marker paths.

## Live smoke

The marker-only implementation was exercised through an isolated local instance while reusing one `session-id`. Successful round trips included:

- `gpt-5.6-sol → claude-opus-5 → gpt-5.6-sol`
- `claude-opus-5 → gpt-5.6-sol → claude-opus-5`
- `gpt-5.6-sol → grok-4.6 → gpt-5.6-sol`
- `gpt-5.6-sol → gpt-5.6-luna → gpt-5.6-sol`

The smoke test logged only status and reasoning presence, not opaque reasoning contents. Native-to-native replay remained accepted; cross-transport requests no longer forwarded an invalid foreign Claude signature.

### DIFF
--- a/src/routes/responses/handler.ts
+++ b/src/routes/responses/handler.ts
@@ -35,6 +35,7 @@ import type {
 import { createResponses as createCopilotResponses } from "~/services/copilot/create-responses"
 
 import { handleResponsesViaMessages } from "./messages-handler"
+import { isMessagesReasoningId } from "./messages-translation"
 import { createStreamIdTracker, fixStreamIds } from "./stream-id-sync"
 import {
   applyResponsesApiContextManagement,
@@ -119,6 +120,7 @@ export const handleResponses = async (c: Context) => {
     responsesTransport,
   )
   if (useMessagesFallback) {
+    filterReasoningForTransport(payload, true)
     return await handleResponsesViaMessages(c, {
       payload,
       publicModel: requestedModel,
@@ -141,6 +143,8 @@ export const handleResponses = async (c: Context) => {
       400,
     )
   }
+
+  filterReasoningForTransport(payload, false)
 
   const recordUsage = createCopilotTokenUsageRecorder({
     endpoint: "responses",
@@ -270,6 +274,18 @@ export const handleResponses = async (c: Context) => {
 
 const isStreamingRequested = (payload: ResponsesPayload): boolean =>
   Boolean(payload.stream)
+
+const filterReasoningForTransport = (
+  payload: ResponsesPayload,
+  useMessagesFallback: boolean,
+): void => {
+  if (!Array.isArray(payload.input)) return
+
+  payload.input = payload.input.filter((item) => {
+    if (item.type !== "reasoning") return true
+    return isMessagesReasoningId(item.id) === useMessagesFallback
+  })
+}
 
 const shouldFallbackToMessages = (
   c: Context,
--- a/src/routes/responses/messages-stream-translation.ts
+++ b/src/routes/responses/messages-stream-translation.ts
@@ -17,6 +17,7 @@ import { CustomToolInputStreamDecoder } from "./custom-tool-input-stream-decoder
 import {
   createMessagesBackedResponsesResult,
   encodeMessagesCompaction,
+  markMessagesReasoningId,
   resolveToolDescriptor,
   ResponsesMessagesTranslationError,
   toResponseId,
@@ -440,7 +441,9 @@ function* startContentBlock(
 
   if (block.type === "thinking") {
     const item: ResponseOutputReasoning = {
-      id: `rs_${state.responseId.slice(-18)}_${event.index}`,
+      id: markMessagesReasoningId(
+        `rs_${state.responseId.slice(-18)}_${event.index}`,
+      ),
       type: "reasoning",
       status: "in_progress",
       summary: [],
--- a/src/routes/responses/messages-translation.ts
+++ b/src/routes/responses/messages-translation.ts
@@ -34,6 +34,13 @@ import type {
 } from "~/lib/types/responses"
 
 export const MESSAGES_COMPACTION_PREFIX = "copilot-api:messages-compaction:v1:"
+const MESSAGES_REASONING_ID_SUFFIX = "__a1"
+
+export const markMessagesReasoningId = (id: string): string =>
+  `${id}${MESSAGES_REASONING_ID_SUFFIX}`
+
+export const isMessagesReasoningId = (id: unknown): boolean =>
+  typeof id === "string" && id.endsWith(MESSAGES_REASONING_ID_SUFFIX)
 
 export const MESSAGES_COMPACTION_PROMPT = [
   "You are performing a CONTEXT CHECKPOINT COMPACTION. Create a handoff summary for another LLM that will resume the task.",
@@ -940,7 +947,9 @@ function translateAssistantOutput(
   for (const [index, block] of response.content.entries()) {
     if (block.type === "thinking") {
       output.push({
-        id: `rs_${createStableHash(`${response.id}:${index}:reasoning`)}`,
+        id: markMessagesReasoningId(
+          `rs_${createStableHash(`${response.id}:${index}:reasoning`)}`,
+        ),
         type: "reasoning",
         status: "completed",
         ...(block.thinking && block.thinking !== "Thinking..." ?
--- a/tests/responses-handler.test.ts
+++ b/tests/responses-handler.test.ts
@@ -1,7 +1,11 @@
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test"
 import { Hono, type Context } from "hono"
 
-import type { AnthropicMessagesPayload } from "~/lib/types/anthropic"
+import type {
+  AnthropicMessagesPayload,
+  AnthropicResponse,
+} from "~/lib/types/anthropic"
+import type { ResponsesResult } from "~/lib/types/responses"
 import type { CompletionPayloadOptions } from "~/routes/messages/handler"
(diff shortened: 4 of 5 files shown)


LaChimere:
@caozhiyuan Please kindly help review this PR for fixing the issue attached. Thanks.

caozhiyuan:
The change is too significant; switching models will break the cache, and cache misses are very expensive @LaChimere

LaChimere:
Thanks — the previous revision was too broad, and the warmup change could increase premium-model usage.

I replaced it with a single focused commit (`f671967`):

- preserves the existing Claude Code small-model warmup behavior
- leaves direct `/v1/messages`, global logging, and Messages terminal SSE shape untouched
- restores same-target reasoning ID/signature byte-for-byte before upstream serialization
- drops reasoning only when provider, actual protocol, or resolved model differs
- documents the one-time legacy-session cache miss instead of hiding it
- reduces the PR from 18 changed files to 8

The same-target route tests inspect the exact payload received by the mocked upstream. A real model change is already a cache divergence (`model_changed` in [Anthropic's cache diagnostics](https://platform.claude.com/docs/en/build-with-claude/cache-diagnostics)); retaining foreign reasoning would not preserve that cache and can make the request invalid.

Current validation: 769 tests, typecheck, lint, build, diff check, and 98% line coverage for the carrier module.

caozhiyuan:
@LaChimere  Thanks again for working on this. After looking at the Codex replay behavior more closely, I think this can be fixed with a much smaller change.

Native GPT reasoning already uses a normal valid `rs_*` item ID, so I would leave GPT Responses output completely untouched.

For Claude reasoning synthesized by our Messages → Responses translation, we can add a small reserved marker to the synthetic reasoning ID while keeping the normal `rs_*` shape, for example:

```text
rs_-a985-1c64c37d0521_0__a1
```

The exact suffix is not important; it only needs to be something explicitly generated and recognized by us.

Then, before forwarding replayed history upstream:

* GPT target:

  * keep unmarked native `rs_*` reasoning
  * drop reasoning carrying our Anthropic marker

* Claude target:

  * keep reasoning carrying our Anthropic marker
  * drop unmarked native GPT reasoning

The filtering should only apply to the current upstream request. It should not permanently mutate the client/history state, so round-trip switching still works:

```text
GPT → Claude → GPT
Claude → GPT → Claude
```

When switching back, the original model can receive its own previous reasoning again.

I would also keep the opaque reasoning data completely unchanged:

* no wrapping or rewriting of `encrypted_content`
* no rewriting of native GPT reasoning IDs
* no general provider/model carrier

Legacy Claude reasoning created before the marker exists cannot be reliably distinguished from native GPT reasoning. I think accepting a one-time loss of reasoning continuity for old sessions is preferable to introducing a much larger provenance framework.

So my preference would be to reduce this PR to:

1. mark only synthetic Claude reasoning IDs,
2. filter incompatible reasoning at the `/v1/responses` upstream boundary,
3. add focused GPT → Claude, Claude → GPT, and round-trip switching tests.

I would avoid changing global logging, provider proxy response rewriting, digest/ETag handling, warmup/model routing, or other unrelated paths unless a focused failing test shows they are actually required.

There is also a separate item-ID issue I noticed: some `custom_tool_call` paths currently appear to use `fc_*`, while the expected convention is `ctc_*` for custom tool calls and `fc_*` for normal function calls. I would keep that separate from this reasoning fix.

If I get some time, I may also implement the smaller version myself.

LaChimere:
Follow-up: I reran the replacement commit itself against an isolated local instance.

- one `session-id` was reused for the entire run
- source/target models: `gpt-5.6-sol`, `claude-opus-5`, `grok-4.6`
- all three sources returned real encrypted reasoning
- all 9 ordered model switches passed in that same session
- upstream-reported cached tokens: GPT 0, Claude 695–713, Grok 1792

The cached-token values are observations, not assertions of a guaranteed cache hit. The important regression check is that same-target carriers were restored and every cross-target request completed without replaying an invalid foreign signature.

caozhiyuan:
When you used 10w tokens, switching models will definitely not hit the cache the first time.  @LaChimere

LaChimere:
Thanks, that makes sense. I agree that switching models in a large session, such as one with ~100k tokens, may cause an expensive cache miss.

I still think smooth model switching is useful, as long as this tradeoff is clear to users. Cache efficiency remains important, but misses may be unavoidable when switching models.

I’ve narrowed the PR to the marker-only approach you suggested. It does not try to preserve cache hits across models—it only prevents incompatible reasoning metadata from making the switched request fail, while keeping the visible conversation history intact.

The updated version passes 75 focused tests and the full 762-test suite. I also verified GPT → Claude → GPT in the same session using the actual prior response output.

Could you please take another look when convenient?

PR #381, opened 2026-09-04
feat: route image requests through configured providers

Falven:
## Summary

Add an optional `imageModel` setting for routing Codex image generation and edit requests through an existing OpenAI-compatible provider.

Codex Desktop exposes image generation as an `image_gen` namespace, but GitHub Copilot reserves that name and rejects it when forwarded unchanged. When `imageModel` is configured, the proxy temporarily aliases that namespace for the Copilot Responses request and restores it in the response. Codex can then execute its built-in image tool, which calls the top-level image endpoint routed by this change.

The setting uses the existing `provider/model` format. The proxy resolves that provider, replaces the request model with the configured model or deployment name, and forwards the request through the provider image path. When `imageModel` is not configured, the existing compatibility filtering and Codex image backend remain unchanged.

Generation requests retain their JSON payload, edit requests retain multipart fields and file data, and both paths preserve query parameters. Provider-scoped image routes are unchanged.

## Configuration

```json
{
  "imageModel": "foundry/my-gpt-image-2-deployment",
  "providers": {
    "foundry": {
      "type": "openai-compatible",
      "baseUrl": "https://<resource-name>.openai.azure.com/openai",
      "apiKey": "<azure-openai-api-key>",
      "authType": "authorization"
    }
  }
}
```

The provider base URL omits `/v1`. The model portion of `imageModel` is the Foundry deployment name.

## Changes

- Add the top-level `imageModel` configuration field.
- Bridge Codex Desktop's reserved `image_gen` namespace through Copilot Responses when image routing is configured.
- Restore the original namespace in streaming and non-streaming responses.
- Route configured generation and edit requests through the existing provider image forwarding path.
- Rewrite the incoming model to the configured provider model.
- Keep the existing Codex behavior when `imageModel` is absent.
- Document Foundry image routing in both READMEs.
- Cover namespace bridging, JSON generation, and multipart edit forwarding.

## Testing

- `bun test` (767 tests passed)
- `bun run typecheck`
- `bun run lint`
- `bun run build`
- Live GHCP Responses request using the aliased namespace
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -703,6 +703,7 @@ Gateway API keys live under `auth.apiKeys` in `config.json`. Manage them with `c
 - **auth.apiKeys:** API keys used for request authentication on non-admin routes. Supports multiple keys for rotation. Requests can authenticate with either `x-api-key: <key>` or `Authorization: Bearer <key>`. If empty or omitted, authentication for non-admin routes is disabled.
 - **auth.adminApiKey:** Single admin key used only for `/admin/*` routes. If missing, the server generates a random key at startup and writes it back to `config.json`. Requests use the same `x-api-key` or `Authorization: Bearer` headers, but regular `auth.apiKeys` never grant access to `/admin/*`.
 - **modelMappings:** Exact `sourceModel -> targetModel` rewrites shared by top-level `POST /v1/messages`, `POST /v1/messages/count_tokens`, `POST /v1/responses`, and `POST /v1/chat/completions` requests. Omit it or leave it as `{}` to disable rewrites. Both the source and target must be non-empty strings. Targets can be regular model IDs or `provider/model` aliases such as `dashscope/qwen3.6-plus`, and the rewrite happens before provider alias parsing. These mappings are not split per interface. The admin endpoints `GET/POST /admin/config/model-mappings` read and update only this field.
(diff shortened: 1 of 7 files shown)


caozhiyuan:
@Falven Can AI use this model to make calls through agents.md or skill.md within codex?

caozhiyuan:
Do not modify the content inside responses

caozhiyuan:
review result:
reason=request.formData() fully materializes the uploaded content before forwarding, breaking the original streaming backpressure. With a 45 MiB multipart request, RSS increases by 94,752,768 bytes; concurrent legitimate image editing requests could lead to significant memory pressure or even OOM.; location=src/routes/images/route.ts:95; fix=use streaming multipart parsing and re-encoding with size limits, only replacing the model field; at minimum, enforce total request size and file count limits before parsing and return 413 when exceeded.; occurrences=1

caozhiyuan:
Will support through model mapping

caozhiyuan:
You can try this, is it okay? https://github.com/caozhiyuan/copilot-api/commit/8d68ba3b641a9d18c0b75ad543c0aa099d5d2ca3 @Falven

PR #382, opened 2026-09-04
feat: support Microsoft Entra auth for configured providers

Falven:
## Summary

Add an `azure-entra` provider auth type for Azure OpenAI endpoints that use Microsoft Entra ID instead of a stored API key.

The shared provider resolver obtains an access token through Azure Identity's `DefaultAzureCredential` using the `https://cognitiveservices.azure.com/.default` scope. Existing provider forwarding then sends the token as `Authorization: Bearer`.

Because authentication happens in the shared resolver, it applies to every configured-provider route, including image generation and edits. Existing API key and Codex OAuth behavior remains unchanged.

## Configuration

```json
{
  "providers": {
    "foundry": {
      "type": "openai-compatible",
      "baseUrl": "https://<resource-name>.openai.azure.com/openai",
      "authType": "azure-entra"
    }
  }
}
```

No `apiKey` is required. `DefaultAzureCredential` can use an Azure CLI session, managed identity, or the standard Azure service principal environment variables.

CLI and desktop provider setup remain unchanged. Entra authentication is configured directly in `config.json`.

## Changes

- Add `azure-entra` to the supported provider auth types.
- Skip API key validation for Entra-authenticated providers.
- Resolve and refresh tokens through `@azure/identity`.
- Reuse the existing bearer authorization path.
- Document the configuration in both READMEs.
- Add coverage for resolving an Entra-authenticated provider.

## Testing

- `bun test` (767 tests passed)
- `bun run typecheck`
- `bun run lint`
- `bun run build`
- `bun run build:desktop`


chrisoro:
Nice, I wanted to add the same :)

caozhiyuan:
## Code Review

**Summary**: Adds an `azure-entra` provider auth type. The shared resolver obtains a token via `DefaultAzureCredential` (scope `https://cognitiveservices.azure.com/.default`) and reuses the existing `Authorization: Bearer` forwarding path, so it covers messages/chat/responses/models/alpha-search/images routes. Existing API key and Codex OAuth behavior is unchanged.

**Context checked**: full diff; PR head checked out locally; read `provider-resolver.ts`, `provider-config.ts`, `config-store.ts`, `provider-proxy.ts` (verified `azure-entra` falls into the Bearer branch), `auth.ts` (CLI prompt unchanged, as described), `models/route.ts` (aggregation catches per-provider errors and degrades to fallback models), `messages/handler.ts`, `server.ts`; inspected all `resolveProviderConfig` callers; verified token caching in the installed `@azure/identity@4.13.2` source (`getBearerTokenProvider` → `bearerTokenAuthenticationPolicy` → `createTokenCycler`: tokens are cached and refreshed with single-flight, so there is no per-request token acquisition).

**Ran locally on the PR head**: `bun test` (767 pass / 0 fail), `bun test tests/provider-resolver.test.ts` (8 pass), `bun run typecheck`, `bun run lint` — all green. I also measured the `@azure/identity` import cost directly.

### Findings

**1. [High priority] Eager top-level `@azure/identity` import adds ~0.5-0.6s to every process startup** — `src/lib/provider-resolver.ts:1`

`provider-resolver.ts` statically imports `@azure/identity` and constructs `DefaultAzureCredential` at module scope. Since the resolver is loaded by the server/desktop startup chain and every provider route, all users pay this cost on every process start even if they never use `azure-entra` — measured at ~485-595ms under Bun on my machine (the whole msal/`open` dependency chain). It also stacks onto every subprocess spawned by the resolver tests.

Fix: lazily initialize inside the `azure-entra` branch — on first hit, `await import("@azure/identity")` and cache the `getBearerTokenProvider` singleton. Important: do not construct a new `DefaultAzureCredential` per call, or the token cycler cache is lost and every request would re-acquire a token (e.g. shelling out to `az` each time).

**2. [Reference] Entra token failure surfaces as an opaque 500** — `src/lib/provider-resolver.ts:65`

If token acquisition fails (expired `az` session, no managed identity, etc.), `resolveProviderConfig` rejects. Handlers like `handleProviderMessagesForProvider` call it before their try blocks, and top-level alias routes go through `ensureConfiguredProviderModelAlias`, so the client gets Hono's default `Internal Server Error` with no hint that upstream auth failed. Consider wrapping Azure credential errors with the provider name and cause, or mapping them to a structured 401/502 in the routes.

**3. [Reference] Per-model type override resets authType** — `src/routes/provider/messages/handler.ts:213`

The override path calls `resolveProviderAuthType(name, undefined, effectiveType)`, resetting authType to the effective type's default. If an `azure-entra` provider sets `models.*.type: "anthropic"` for a model, the Entra token is sent as an `x-api-key` header and upstream will 401. This is a pre-existing pattern, but `azure-entra` makes it newly reachable. Consider preserving the configured authType in the override path, or documenting that `azure-entra` should not be mixed with per-model type overrides.

**Release risk**: low-medium. The core path is verified: token caching/auto-refresh is handled by the SDK token cycler, Bearer forwarding reuses the existing path, the models aggregation route has per-provider fallback, and existing auth paths are untouched. The main regression is the ~0.5s startup overhead for all users (verified; worth fixing before merge). The failure-mode opacity and the override edge case are low-probability scenarios.

Review Decision: BLOCKING (Findings Total: 3 — CRITICAL=0, HIGH_PRIORITY=1, REFERENCE=2)

caozhiyuan:
@Falven

PR #383, opened 2026-09-04
Add usage viewer statistics periods

chrisoro:
Adds calendar-based and lifetime usage periods to the localhost usage viewer and token-usage API.

I wanted to keep your current logic as is, so I just added more options without changing too much. I think there is lots of repetition/duplication in the original code, so you might want to refactor that part.

- Adds Week to date, Month to date, and Lifetime viewer options.
- Preserves existing Day, rolling Week, and rolling Month periods.
- Keeps Day as the default period.
- Adds API support for all 3 options.
- Week to date starts Monday at 00:00.
- Month to date starts on the first day at 00:00.
- Lifetime starts at the earliest recorded event.

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,16 +585,16 @@ The dashboard provides a user-friendly interface to view your Copilot usage data
 
 - **API Endpoint URL**: The dashboard is pre-configured to fetch data from your local server endpoint via a URL query parameter. You can manually switch this to any other compatible API endpoint.
 - **API Key Authentication**: If API Key authentication is enabled, enter a raw API key (sent as the `x-api-key` header) or `Authorization: Bearer <key>`. Credentials are remembered in the browser's local storage per endpoint origin, and switching to a different endpoint origin does not automatically send the previous credential.
-- **Period Selector**: Choose from Day, Week, or Month time ranges. The URL query parameter updates automatically when you switch, making it easy to bookmark and share.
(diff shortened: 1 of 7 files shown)


caozhiyuan:
Thanks for the contribution. I reviewed the full change together with the surrounding storage, routing, viewer, authentication, SQLite, test, and CI context. I did not find any blocking correctness, security, or compatibility issues. The three new periods are wired consistently through the summary, daily, and events APIs, and the boundary behavior is well covered by tests.

Two non-blocking follow-ups:

1. Lifetime daily aggregation scales linearly with the age of the database. [`getTokenUsageDailySummary`](https://github.com/caozhiyuan/copilot-api/blob/ebeeefcddd36b474eb2f6ceb646f9f48545dc1e0/src/lib/token-usage/store.ts#L912-L918) creates one bucket per calendar day and calls `getModelSummaries` for every bucket, which performs two SQLite aggregate queries per day. The viewer also fetches all daily buckets and renders SVG points for each one. On Bun 1.4.0 with a sparse database, a 5-year range produced 1,827 buckets in about 0.52 seconds and a 470 KB response; a 10-year range produced 3,653 buckets in about 1.49 seconds and a 940 KB response. Consider querying all date/model aggregates with a single `GROUP BY`, then filling empty buckets in memory, and using monthly aggregation or controlled downsampling for the Lifetime chart.

2. The usage-viewer documentation is now stale. [`README.md`](https://github.com/caozhiyuan/copilot-api/blob/ebeeefcddd36b474eb2f6ceb646f9f48545dc1e0/README.md#L588-L597) still says the selector only supports Day, Week, and Month and describes the trend chart as Week/Month only. It would be useful to document all six period values, their range semantics, and an updated URL example.

Validation performed: 26 targeted tests passed, TypeScript checking passed, the changed TypeScript files passed ESLint, and the viewer's inline script compiled successfully. The PR CI check is also green.

Overall review decision: **non-blocking**.

chrisoro:
Thanks for the review, I'll do the improvements when I'm back home

PR #391, opened 2026-09-08
feat: 强化容器运行、Compose 部署与迁移保护

Thinker-Joe:
## Summary / 变更摘要

关联的独立发布策略提案：#392。两个 PR 可以分别接受或拒绝。

希望将容器运行层面的改进贡献给上游。本 PR 只涉及容器运行强化、通用 Compose、迁移文档和容器测试；**不修改现有镜像发布工作流，也不改变开发镜像或 `latest` 策略**。发布策略会通过另一个独立 PR 征求维护者意见，可以只合并本 PR。

- 镜像以非 root 用户运行，应用代码保持 root 所有，持久化目录统一为 `/data`；入口脚本在目录不可写或已有配置不可读时明确退出，避免静默使用空配置。
- 保留 `--auth` / `auth`、旧 `GH_TOKEN` 和新 Token 环境变量兼容；GitHub Token 不进入进程参数。默认启用代理环境变量处理，并支持 `--no-proxy-env`。
- CLI 端口优先于 `PORT`，健康检查读取实际监听地址，支持动态端口和 IPv6、绕过代理并限制耗时。
- Compose 默认只绑定宿主机回环地址，使用只读根文件系统、临时可写目录、移除 capabilities、禁止提权和日志轮转。新增通用 bind mount 覆盖配置，拒绝自动创建不存在的宿主机目录。
- 补齐中英文首次初始化、备份、权限迁移与回滚说明，以及入口脚本、健康检查、启动和真实容器生命周期测试。独立 Docker CI 在 AMD64 / ARM64 原生构建及运行，不发布镜像。

### 范围与兼容性

基于当前上游 `dev`（`cdcd2cb`）单独建立提交，没有携带 fork 的合并历史、私有部署地址、SV 专属配置或日志。上游已合入的配置保护修复 `5d7ea5b` 及其测试保持原样，不重复提交。

**迁移提示：**旧镜像的 `/root/.local/share/copilot-api` 挂载需迁移到 `/data`；保留原宿主机数据来源、停服备份并修正 UID/GID。已有 bind mount 不能直接切换到基础 Compose 的新命名卷。完整步骤见 `docs/docker.zh-CN.md`，本 PR 不自动修改任何已部署服务器。

## Test Evidence / 验证记录

- [x] `bun run lint:all`
- [x] `bun run typecheck:all`（服务端及桌面）
[...]

PR #393, opened 2026-09-09
feat: harden Docker runtime and document data migration

Thinker-Joe:
## Summary

Make Docker deployments run as a non-root user with an explicit persistent-data location, and provide a documented upgrade path for existing installations. This replaces #391 under a renamed branch; the runtime scope remains separate from release publishing.

## Motivation

The current image runs as root and stores application state under a root-owned home directory. Moving to a non-root runtime requires more than changing `USER`: existing mounts need ownership preparation, authentication commands must use the same state as the server, and health checks must follow the actual listening port.

This PR makes those requirements explicit without adding automatic ownership changes or moving existing data behind the operator's back.

## Changes

- Run as `bun`, retain root-owned application code, and use `/data` for persistent state.
- Add a local-build Compose configuration with loopback-only host publishing, a read-only root filesystem, writable temporary storage, dropped capabilities, no-new-privileges, and bounded logs.
- Keep authentication subcommands and legacy `--auth` usage working. Preserve the existing environment-based GitHub-token interface and avoid adding tokens to process arguments.
- Fail early with actionable diagnostics when the data directory is inaccessible or existing configuration is unreadable. Configuration-preservation behavior already present in the application is retained, not reimplemented.
- Make `PORT` the CLI default while preserving explicit CLI precedence. Publish the bound address for a proxy-independent health probe, including dynamically assigned ports and IPv6.
- Set `XDG_CACHE_HOME=/data/cache` in Compose so VSCode device IDs can persist with a read-only root filesystem.
- Add focused tests, opt-in container lifecycle tests, native AMD64/ARM64 validation, and bilingual deployment and migration documentation.

## Compatibility and migration

**Existing deployments require an explicit migration.** The former container path is `/root/.local/share/copilot-api`; the new default is `/data`. Stop and back up the service, preserve the original data source, and prepare ownership for the image's actual UID/GID before starting the new image.

The base Compose file uses a named volume. Existing bind-mount users must keep their data source through `docker-compose.bind.yaml`; switching to the base file alone selects a different, initially empty volume. The override refuses to create a missing host directory. No automatic data copy, recursive startup-time `chown`, or root fallback is introduced.

A GitHub token still does not replace the gateway API key. Health checks report local liveness, not provider authentication, quota, or upstream availability.

## Scope

This PR does not change image publishing, registry destinations, release tags, or application protocol translation. It can be reviewed and merged independently of the release-workflow PR (#394). Fork-specific deployment settings and development-image publishing are not included.

## Code Review

Self-review using `copilot-api-code-review`:

- **Summary**: Reviewed runtime identity, persistent-state access, CLI compatibility, health publication, Compose restrictions, and migration instructions.
- **Context Checked**: Diff against `dev`; Dockerfile and ignore rules; both Compose files; entrypoint and healthcheck scripts; `src/start.ts`, `src/lib/server-health.ts`, atomic file writing, configuration loading and device-ID cache resolution; the five container/health test files; Docker CI; README and bilingual deployment docs.
- **CHANGESET_SUMMARY**: Introduces a non-root Docker runtime and explicit deployment contract, protects existing state through early permission checks and operator-led migration, and validates startup and health behavior. Publishing policy is unchanged.
- **CODE_REVIEW_SUMMARY**:

```text
Review Decision: NON_BLOCKING
Findings Total: 0 (CRITICAL=0, HIGH_PRIORITY=0, REFERENCE=0)

[CRITICAL]
None

[HIGH_PRIORITY]
None

[REFERENCE]
None
```

- **Test Notes**: Native Linux AMD64 and ARM64 jobs passed, including Compose validation, native image builds, and the three container lifecycle tests. Local Windows results and limitations are listed below.
- **Release Risk**: Low to medium. Native Linux validation covers runtime identity, Compose integration, and container lifecycle behavior. The remaining risk is operator migration: existing deployments must follow the documented ownership and mount-path steps.

## Test evidence

Native Linux CI (fork run [34322112246](https://github.com/Thinker-Joe/copilot-api/actions/runs/34322112246), head `eed36f7`), executed on both `ubuntu-24.04` and `ubuntu-24.04-arm`:

- Repository lint of the container-related TypeScript, `bun run typecheck`, and `bun test`: passed on both architectures.
- `docker compose config` for the base and merged bind-mount configurations: passed on both architectures.
- Native image build (`docker build --tag copilot-api:test .`): passed on both architectures.
- `bun test tests/docker-smoke.test.ts`: **3 passed, 0 failed** on each architecture, covering hardened startup, proxy-safe health probing, CLI port override, container recreation, unwritable-volume failure, and preservation of unreadable legacy data.

The same head also passed the fork's `CI` workflow (run [34322112209](https://github.com/Thinker-Joe/copilot-api/actions/runs/34322112209)): lint, root and desktop typecheck, `bun test` (**902 passed, 0 failed**), and `bun run build`.

Validated locally on Windows with Bun 1.3.14:

- `bun run typecheck:all`: passed (root and desktop).
- `bun test`: **889 passed, 16 skipped, 0 failed**. Skips include POSIX permission/entrypoint cases and the opt-in Docker suite.
- Targeted container/health tests with coverage: **22 passed, 12 skipped, 0 failed**; `src/lib/server-health.ts` reached **100% line and function coverage**. This does not measure shell-script coverage.
- `bun run build` and `bun run build:desktop`: passed.
- YAML parsing and deployment-contract assertions: passed. Shell syntax was checked with Git Bash.
- Standard `bun run lint:all` failed on CRLF formatting in this Windows checkout. Root and desktop lint passed when only Prettier's `endOfLine` was overridden to `auto`, preserving the remaining repository options. No lint configuration was changed.
- Docker is unavailable on the local Windows host, so container execution was verified only through the native Linux CI jobs above. Desktop dependency installation encountered an Electron binary-download failure locally; the CI jobs installed desktop dependencies and passed the desktop typecheck.


caozhiyuan:
Thanks for the work on hardening the Docker runtime. The overall direction (non-root user, `/data`, read-only rootfs, explicit migration docs) looks good, but I would like to ask for simplifications in the spirit of KISS before this moves forward. I am closing this PR for now; please feel free to open a slimmer follow-up.

1. `entrypoint.sh` injects `--proxy-env`, but it is a no-op under Bun. `initProxyFromEnv()` in `src/lib/proxy.ts` returns early when `typeof Bun !== "undefined"` (the undici global dispatcher is only installed on the Node path), and `getProxyEnvDispatcher()` has zero callers in `src/`. Bun's native `fetch` already honors `HTTP_PROXY`/`HTTPS_PROXY` environment variables, so proxying in the container works from the environment alone. Please drop the injected flag from the entrypoint (keeping the CLI flag itself for Node compatibility is fine) and update the `docker-entrypoint.test.ts` expectations accordingly.

2. `docker-test.yml` duplicates the main CI: it re-runs the full `bun test` suite plus typecheck on two runners. Please narrow it to the container/health test files and keep the image build, `compose config` validation, and smoke tests, which are the parts that add unique value.

3. `tests/server-health.test.ts` has two floating `rejects` assertions: the "rejects a server without a listening address" test is neither async nor returned, and the final `expect(fetch(url)).rejects` is not awaited, so regressions there would stay green. Please `await` both.

4. KISS question on the healthcheck design: the publish-file machinery (`server-health.ts`, `healthcheck.sh`, `COPILOT_API_HEALTHCHECK_FILE`, plus related tests and docs) exists to support CLI `--port` overrides inside the container, yet Compose pins the internal port to 4141. If port overrides in containers are not a real use case, the simpler design is a fixed internal port with a static probe (keeping `--noproxy '*'`, which is the part of `healthcheck.sh` genuinely worth keeping). Please either justify keeping the current design or simplify to fixed-port plus static probe. My preference is the simpler one.

Docs nit: "server startup enables proxy environment handling; `--no-proxy-env` disables it" overclaims under Bun, where the flag currently changes nothing. Worth rewording when the above is addressed.

Thanks again, and happy to review the follow-up.

